PR #5, opened 2016-05-11
Allow users config to override the default config.

rwjblue:
Without this change, it is not possible to customize any of the default values (because any customizations are clobbered by the default config).


kellyselden:
Could there be a point release of this? I need it to get coveralls reporting to work. Have to filter out browserify noise.

kategengler:
Published 0.1.2. Sorry, meant to do it earlier but I tend to forget about notifications I see during the workday (the original PR).

Issue #8, opened 2016-05-14
paths in "lcov.info" don't relate to project setup


kellyselden:
After I resolve #7, I saw that the paths in "lcov.info" are not right. I was getting in my coveralls logs `Error: ENOENT: no such file or directory, open '/home/travis/build/kellyselden/package-hint-historic-resolver/package-hint-historic-resolver/adapters/application.js'`. It seems the second "app name" needs to be either "app" or "addon" for the path to be correct for coveralls and maybe other reporters.

Issue #7, opened 2016-05-14
files from addons bleed into coverage results


kellyselden:
Not all the files in the picture are mine, some come from addons:
![capture2](https://cloud.githubusercontent.com/assets/602423/15271149/38a90e64-19f0-11e6-9704-aaff4255089b.PNG)

kategengler:
Are the files that are from addons all from addon's app tree? Those are ones I've observed are mingled. How could I filter them out?

benoror:
Can confirm addons files were instrumented too, which were responsible for the discrepancies. PR with @kellyselden changes concerning to **Addons files exclusion** only: https://github.com/kategengler/ember-cli-code-coverage/pull/10

kategengler:
I still don't think all addon files bleed into coverage, just those that have been merged with the 'app' tree

rwjblue:
Agreed, it should be the addon's `app` tree files that are bleeding in at the moment.

sandersky:
Addressed this in #11 which I believe covers more edge cases than #10

PR #9, opened 2016-05-14
WIP: add external coverage support (coveralls)

kellyselden:
Resolves both #7 and #8, although probably not the right implementation. It involves:
* checking if the file exists, to remove all the files that came in from addons
* changing the path in "lcov.info" so that the paths are "real" enough for coveralls to find them

### DIFF
--- a/lib/coverage-instrumenter.js
+++ b/lib/coverage-instrumenter.js
@@ -2,11 +2,13 @@
 
 var Filter = require('broccoli-filter');
 var Istanbul = require('istanbul');
+var pathFixer = require('./path-fixer');
 
 CoverageInstrumenter.prototype = Object.create(Filter.prototype);
 CoverageInstrumenter.prototype.constructor = CoverageInstrumenter;
 function CoverageInstrumenter(inputNode, options) {
   options = options || {};
+  this._app = options.app;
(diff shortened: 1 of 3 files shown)


benoror:
I see discrepancies in https://codecov.io/ coverage results vs **local**, might be related to addons as well??

lib/coverage-instrumenter.js
```diff
@@ -22,6 +24,9 @@ CoverageInstrumenter.prototype.processString = function(content, relativePath) {
     noAutoWrap: true
   });
 
+  // fixes the path so the lcov.info can be read by 3rd parties (coveralls)
+  relativePath = pathFixer.fixPath(relativePath, this._app);
```
benoror: in my use-case I fixed the paths directly with my provider: https://github.com/codecov/support/blob/master/site/docs.md#-fixing-paths

sandersky:
I believe my implementation in pull request #11 will give you what you desire. I have tested against an app locally that uses ES6 and ES7 decorators. The coverage being generated is filtering out files from addons, reporting proper file paths, as well as providing line counts against the original ES6 source code rather than the transpiled E56 code.

PR #10, opened 2016-05-15
exclude addons

benoror:
cherry picked from https://github.com/kategengler/ember-cli-code-coverage/pull/9


sandersky:
I've implemented the same functionality in #11 but I believe my solution handles more edge cases. I've verified it works with both apps and addons and also reports coverage for code living within an addons dummy app.

sandersky:
@benoror I believe you can close this PR out now as the functionality landed today in another PR.

benoror:
Should be fixed in https://github.com/kategengler/ember-cli-code-coverage/pull/11

PR #11, opened 2016-05-18
Filter out files from addons, fix file paths, and rename config file

### LINKED ISSUE #4: Change config file name?
`config/coverage-config.js` seems redundant, what do you think about changing the file to `config/coverage.js` or `config/code-coverage.js`?

### LINKED ISSUE #7: files from addons bleed into coverage results

### LINKED ISSUE #8: paths in "lcov.info" don't relate to project setup

sandersky:
* **Filter Out Files from Addons**
The code coverage currently being reported by this addon includes files from Ember addons the consuming app/addon depends on, which make the metrics a little bit misleading. This pull request filters out those files so the coverage reported is only for files from the consumers app/addon. Resolves #7.

* **Fix File Paths**
Currently the file paths being reported in coverage look something like `foo/components/bar.js` which is actually referencing the file at `app/components/bar.js`. Since the file paths being reported are not real file paths lcov viewers do not properly map the lcov info to the corresponding files, making it somewhat hard to utilize. This pull request fixes the file paths so that they reference real files, making editor plugins such as `lcov-info` for Atom highlight coverage in files. Resolves #8.

* **Rename Config File**
Renamed the configuration file from `config/coverage-config.js` to `config/coverage.js`. Resolves #4.


sandersky:
@kategengler would love to get this reviewed/merged as soon as possible so my team can start switching over to `ember-cli-code-coverage` from `ember-cli-blanket`.

index.js
```diff
@@ -19,15 +92,24 @@ module.exports = {
...
-      var instrumentedNode = new CoverageInstrumenter(appFiles, {annotation: 'Instrumenting for code coverage'});
-      return new BroccoliMergeTrees([tree, instrumentedNode], { overwrite: true });
+  preprocessTree: function(type, tree) {
+    if (!this._coverageIsEnabled() || type !== 'js') {
+      return tree;
     }
-    return tree;
+
+    var packageFilePath = path.join(this.project.root, 'package.json');
+    var appName = JSON.parse(fs.readFileSync(packageFilePath)).name;
```
kellyselden: You can probably get this via a property on `this`.

sandersky: Agreed, committed a change to use `this.parent.pkg.name` instead.

lib/babel-istanbul-instrumenter.js
```diff
@@ -0,0 +1,349 @@
+/* global Symbol */
```
kellyselden: Did you write this file, or is copied from somewhere?

sandersky: I got it from a co-workers fork here: https://github.com/sophypal/ember-cli-code-coverage. @sophypal was this copied from *isparta*?

sandersky: I added a comment as to where the code came from and did some minor cleanup on it.

sophypal: This was from `isparta` which is not maintained anymore.  I actually meant to clean up my fork so it doesn't look like it came straight out of `babel`.

kellyselden:
This PR is a behemoth, and I think the first step to getting it all merged is to break it into pieces. For all those issues linked, could you please create a separate PR?

patcoll:
This PR is indeed large. However, it seems stable. I am adapting it in my app right now without problems.

sglanzer:
@kellyselden I don't think it makes a lot of sense to make @sandersky break this into separate PRs to get this reviewed/merged unless there are perceived issues or controversy with portions of the PR.

He's done a ton of work to add a bunch of useful features to this addon, why not make it easier on him and review the current PR?

rwjblue:
On the one hand it is pretty annoying to split up this PR, but on the other it is harder to review it and understand the various changes needed.  

I'm going to try and review this PR to provide actionable feedback either way...

index.js
```diff
@@ -12,36 +12,176 @@ var config = require('./lib/config');
...
+
+  /**
+   * Check if a template file exists within the current app/addon
+   * Note: Template files are already compiled into JavaScript files so we must
+   * check for the pre-compiled .hbs file
+   * @param {String} relativePath - path to file within current app/addon
+   * @returns {Boolean} whether or not the file exists within the current app/addon
+   */
+  _doesTemplateFileExist: function(relativePath) {
+    relativePath = relativePath.replace('.js', '.hbs');
```
rwjblue: We should support any known template extension.  You can get these extensions via:

```js
    var templateExtensions = this.registry.extensionsForType('template');
```

(stolen from https://github.com/ember-cli/ember-cli/blob/d987b4f7dd1e8b2476193b47ac0e5e489cb7f585/lib/models/addon.js#L605)

index.js
```diff
@@ -12,36 +12,176 @@ var config = require('./lib/config');
...
+  /**
+   * Check if a file exists within the current app/addon
+   * @param {String} relativePath - path to file within current app/addon
+   * @returns {Boolean} whether or not the file exists within the current app/addon
+   */
+  _doesFileExistInCurrentProject: function(relativePath) {
+    relativePath = path.join('app', relativePath);
+
+    try {
+      fs.statSync(relativePath);
```
rwjblue: I'd prefer to use `exists-sync` package to `try`/`catch`.

index.js
```diff
@@ -12,36 +12,176 @@ var config = require('./lib/config');
...
+  /**
+   * Check if a file exists within the dummy app
+   * @param {String} relativePath - path to file within dummy app
+   * @returns {Boolean} whether or not the file exists within the dummy app
+   */
+  _doesFileExistInDummyApp: function(relativePath) {
+    relativePath = path.join('tests', 'dummy', 'app', relativePath);
+
+    try {
+      fs.statSync(relativePath);
```
rwjblue: I'd prefer to use `exists-sync` package to `try`/`catch`.

index.js
```diff
@@ -12,36 +12,176 @@ var config = require('./lib/config');
...
+   * @returns {[type]} processed tree
+   */
+  preprocessTree: function(type, tree) {
+    // If coverage isn't enabled or tree is not JavaScript tree then we don't need to alter the tree
+    if (!this._isCoverageEnabled() || type !== 'js') {
+      return tree;
     }
-    return tree;
+
+    // Make sure we exlcude files defined in the configuration as well as files from addons
```
rwjblue: typo `exlcude` -> `exclude`

index.js
```diff
@@ -12,36 +12,176 @@ var config = require('./lib/config');
...
+    excludes.push(this._filterOutAddonFiles.bind(this, name));
+    return excludes;
+  },
+
+  /**
+   * Determine whether or not coverage is enabled
+   * @returns {Boolean} whether or not coverage is enabled
+   */
+  _isCoverageEnabled: function() {
+    var value = process.env[this._getConfig().coverageEnvVar];
```
rwjblue: Why not go with `return !!process.env[this._getConfig().coverageEnvVar];`?  What does the extra complexity here gain us?  Is this just to support setting `COVERAGE=false`?  If that is it, maybe something like this would be simpler:

```js
var value  = process.env[this._getConfig().coverageEnvVar];

return value && value !== 'false';
```

Or (if `true` is the magic value):

```js
return process.env[this._getConfig().coverageEnvVar] + '' === 'true';
```

lib/babel-istanbul-instrumenter.js
```diff
@@ -0,0 +1,256 @@
+/**
```
rwjblue: If we are going to check this file into the repo in ES5, and isparta is unmaintained we should probably just rewrite this in idiomatic ES5 style instead of using the babel transpiled version of the isparta code.

Specifically, things like the following make life a little harder:

```js
var _this = _possibleConstructorReturn(this, Object.getPrototypeOf(Instrumenter).call(this));
//...snip...
var result = this._r = (0, babelTransform)(code, extend({}, this.babelOptions, { filename: fileName }));
```

rwjblue: This file is pretty complicated, and ultimately makes the maintenance of this repo much harder (at least it adds significantly to the cognitive overhead required to fix/change things).  Maybe we should start out with the other changes, and circle back around to this one if it is absolutely critical...

lib/coverage-instrumenter.js
```diff
@@ -16,12 +56,18 @@ CoverageInstrumenter.prototype.extensions = ['js'];
 CoverageInstrumenter.prototype.targetExtension = 'js';
 
 CoverageInstrumenter.prototype.processString = function(content, relativePath) {
-  var instrumenter = new Istanbul.Instrumenter({
+  var instrumenter = new Instrumenter({
+    babel: {
+      blacklist: ['es6.modules'],
+      nonStandard: false,
+      optional: ['es7.decorators']
```
rwjblue: This seems odd to hard code. Ultimately, we need to pass through whatever the app being compiled is using for its plugins...

lib/coverage-instrumenter.js
```diff
@@ -1,12 +1,52 @@
 'use strict';
 
 var Filter = require('broccoli-filter');
-var Istanbul = require('istanbul');
+var Instrumenter = require('./babel-istanbul-instrumenter');
+var fs = require('fs');
+var path = require('path');
+
+function getPathForRealFile(relativePath, root) {
```
rwjblue: This should share the logic from `index.js`, where we are doing a very similar thing.

lib/coverage-instrumenter.js
```diff
@@ -1,12 +1,52 @@
...
+  try {
```
rwjblue: we should be able to avoid `try`/`catch` here...

lib/coverage-instrumenter.js
```diff
@@ -1,12 +1,52 @@
...
+    fs.statSync(path.join(root, relativePath))
+    return relativePath
+  } catch (err) {
```
rwjblue: we should be able to avoid nested `try`/`catch` here...

lib/coverage-instrumenter.js
```diff
@@ -1,12 +1,52 @@
...
+    try {
+      var templatePath = relativePath.replace('.js', '.hbs')
```
rwjblue: Should account for alternate template file types

tests/.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
```
rwjblue: I'd like this to be `.eslintrc.js`

rwjblue:
OK, I did a round of review, please reply back inline as needed to clarify or confirm my comments.

The changes needed based on my feedback are fairly minor, so perhaps it does make sense to try to fix this up and land it directly instead of separate PR's.

I will try to circle around with @kategengler on it...

sandersky:
@rwjblue I believe I've addressed your concerns. Would love additional feedback when you get a chance.

patcoll:
I am anxious to see a final review + merge for this PR

rwjblue:
Will try to review tonight, thanks for your patience...

rwwagner90:
+1 this looks dope. Glad someone finally got Istanbul up and running!

lib/babel-istanbul-instrumenter.js
rwjblue: I'd like to land this PR without this file.  It adds quite a bit of extra weight maintenance wise, and it isn't clear to me how much value it adds.

Can you remove and update?  I'm happy to work through the specific pros/cons in a separate PR after this has landed, but I really don't want to block the other refactors while we hash that out...

rwjblue:
Looks great, only have that one blocker (moving the `lib/babel-istanbul-instrumenter.js` usage into a follow up PR).

sglanzer:
@sandersky ^

hbdahl:
This does definitely look amazing! It would be astronomically beneficial to the community if the updated code gets merged! 👍

lib/babel-istanbul-instrumenter.js
sandersky: The value add of this is that the code coverage reports proper line numbers on transpiled code. Without this piece the line numbers are wrong and that makes code coverage metrics a lot less useful.

rwjblue: Sounds like we will have a grand discussion in the follow up PR then. 😸

sglanzer: Absolutely zero offense meant by this, but it seems as though your team may not have enough time available currently to move this project forward with pace.  This is a slight problem for us as I've essentially stopped maintenance of the blanket project in anticipation of this and our company paused all work on ember-cli-istanbul as well.

It may be prudent for us to resume work on ember-cli-istanbul at this stage as we have a minimum of 4 engineers who could contribute to the project immediately.  We would also be happy to grant collaborator status to members of the ember-cli team.

Let me know if this course would be overly controversial but I would rather have a useable, if slightly flawed, product that we iterate on than remain in limbo.

rwjblue: No offense taken, but my requests here seem reasonable (perhaps only to me?) and have been communicated very clearly (for most of the time this PR has been waiting). As I mentioned, a number of times before, once this is addressed, the bulk of the work will be merged and we can have a specific conversation about this file on its own merit.

index.js
rwjblue: @sandersky - Has this been addressed?

lib/babel-istanbul-instrumenter.js
sglanzer: Fair enough, I'll ask @sandersky to split out the file, having spoken with him yesterday the concern was that:

1. It's a reasonable inclusion given that line numbers don't map correctly otherwise
2. The initial review took around 2 weeks to receive a response

As long as the second PR doesn't get stuck then there really isn't an issue, so we'll split it and go from there.

Thanks for the response - here's to a complete code coverage solution :)

sandersky:
@rwjblue pulled out that chunk of code for a future PR.

rwjblue:
@sandersky - Thank you!!!  Sorry for taking so long on this, but I really appreciate the hard work you have put in here.

Looking forward to that follow up PR...

lib/babel-istanbul-instrumenter.js
sophypal: Just to be clear, having this without `babel-istanbul-instrumenter` will add a whole lot of value by itself. Since we're now using the `preprocessTree` hook and `esprima` can inherently parse ES6, we get line numbers that map correct for ES6. The trouble comes only with ES7 features that we have gotten use to using like `decorators`.

sandersky: @sophypal thanks for pointing that out. I can't use this yet because of my use of `ember-computed-decorators` in every app I work on. But I agree the value add for ES6 is definitely there.

rwjblue:
```
% npm publish
+ ember-cli-code-coverage@0.2.0
```

sglanzer:
Awesome :)

PR #13, opened 2016-05-31
Add JSDoc and perform minor cleanup

sandersky:
@kellyselden here is the first smaller PR for my behemoth PR. This one doesn't change anything functionally as it is just adding JSDoc and performing some minor cleanup.


.travis.yml
```diff
@@ -8,10 +8,6 @@ node_js:
 
 sudo: false
 
-cache:
```
kellyselden: caching node_modules is good for testing times. why remove?

sandersky: I've ran into many issues in the past with caching node_modules with Travis where it ends up getting cached in a bad state. When this happens it is not obvious that clearing the cache will resolve issues and many hours end up being wasted digging into the issue.

rwjblue: @sandersky - I have also felt this pain and definitely understand the goal/intent here, but I think this addon should try to stick as close to the default addon blueprint as possible.  Straying from that makes future upgrades harder...

tests/.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
```
rwjblue: Can you rename this to `.eslintrc.js` and `module.export` this?  Using JS files will allow us to cascade rules a bit easier...

rwjblue:
I left two minor comments inline, but other than those this looks great!

sandersky:
Thanks @rwjblue, I have addressed your concerns.

PR #24, opened 2016-08-03
Avoid requiring jQuery for sending the coverage report back.

rwjblue:
Works the same:

screenshot

And doesn't troll apps running without jQuery (don't ask yet if that is
a thing plz :wink: ).

### DIFF
--- a/lib/templates/test-body-footer.html
+++ b/lib/templates/test-body-footer.html
@@ -3,37 +3,37 @@
     var coverageData = window.__coverage__;
     var data = JSON.stringify(coverageData || {});
 
-    $.ajax({
-      type: 'POST',
-      async: false,
-      url: '/write-coverage',
-      datatype: 'json',
-      contentType:'application/json; charset=utf-8',
-      data: data,
-      error: function(jqXHR, textStatus, errorThrown ) {
-        throw errorThrown;
-      },
-      success: function(data) {
-        if (!data || !data.total) {
-          return;
-        }
+    var request = new XMLHttpRequest();
+    request.open('POST', '/write-coverage', true);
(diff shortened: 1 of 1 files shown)


rwjblue:
tests are failing because `ember test` is timing out, I'm not sure why yet.

When using this `npm link`'ed into a test ember-cli project it works properly...

kategengler:
It looks like Phantom isn't happy with this. I have no patience for that so I went ahead and got rid of Phantom from the testem config and added some more output in the case of failure so future failures will be less mysterious (https://github.com/kategengler/ember-cli-code-coverage/commit/dc9c95f2f0a8c80488e8e198cc7f01fbb4040a24)

lib/templates/test-body-footer.html
```diff
@@ -3,37 +3,35 @@
...
+          return name + ' ' + data.total[name.toLowerCase()].pct + '%'
+        });
+
+      var resultsText = document.createTextNode(results.join(' | '));
+      var element = document.createElement('div');
+      element.style = 'background-color: white; color: black; border: 2px solid black; padding: 1em;position: fixed; left: 15px; bottom: 15px';
+      element.appendChild(resultsText);
+      document.body.appendChild(element);
+
+      callback();
```
kategengler: This was wrapped in an `if (callback)` before, for the case when it is run at `/tests` and there is no Testem callback. Not sure how to write a test that could catch that.

rwjblue: Gotcha.  I will update, but I'm also unsure of how to test that since we can't run `ember test`...

kategengler: Yeah, I guess so long as it's there...

PR #38, opened 2016-09-15
add Windows notes

kellyselden:


kellyselden:
I always use `cross-env` when setting env vars, because I switch between mac and windows a lot.

rwjblue:
Thanks!

PR #46, opened 2016-09-21
Handle addons with different package names

### LINKED ISSUE #45: No coverage when addon and package name don't match
We have an npm scope associated with our private registry, so our internal addons all have scoped package names like `@scope/foo-bar`. They use the unscoped version (`foo-bar`) for the name in their `index.js`, which unfortunately causes ember-cli-code-coverage to filter out all our actual addon files.

Locally, I've had success replacing instances of `project.pkg.name` with `project.name()`, which returns the package's "base name" (everything after the last `/`). It works for our use case, and I started to open a PR with that change, but it doesn't seem like it solves the underlying problem — the name in `package.json` and the name in `index.js` aren't necessarily related.

Given that the covered addon is looked up by name [here](https://github.com/kategengler/ember-cli-code-coverage/blob/master/index.js#L20), it seems like kind of a catch-22 unless there's another way to track down the `Addon` instance without knowing its name. Any guidance on this? Are mismatched package and addon names even something this addon needs to handle, or is it enough of an antipattern that it should just be discouraged?

dfreeman:
This should fix #45, where addons whose name is different from the one in their `package.json` don't get coverage.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ module.exports = {
 
   included: function() {
     if (this._isCoverageEnabled() && this.parent.isEmberCLIAddon()) {
-      var coveredAddon = this.project.findAddonByName(this.project.pkg.name);
+      var coveredAddon = this._findCoveredAddon();
       var coverageAddonContext = this;
 
       coveredAddon.processedAddonJsFiles = function (tree){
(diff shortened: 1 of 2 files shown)


dfreeman:
Just checking in — anything I can to do help get this merged? We'd love to start using this in a bunch of our internal addons, but right now #45 is a blocker (short of just using a forked version)

rwjblue:
Looks great! Sorry for the delay here...

PR #50, opened 2016-10-07
Use posix paths for dealing with modules.

### LINKED ISSUE #48: Addon support: no coverage for addon files on Windows
I did not get any coverage information for files within the addon subdirectory. This problem seems to be specific for windows environments. It seems, that the function [here](https://github.com/kategengler/ember-cli-code-coverage/blob/0b6ebaaf245739c255edbd5e2a557008ba11a683/index.js#L132)
```
 _doesFileExistIsCurrentProjectAddonModule: function(relativePath) {
    relativePath = path.join('addon', relativePath.replace(path.join('modules', this.project.pkg.name),''));

    if (this._existsSync(relativePath)) {
      return true;
    }

    return this._doesTemplateFileExist(relativePath);
  },
``` 
in index.js has problems with the path separator on Windows ("/" vs. "\"). Inserting `relativePath = path.join(relativePath);` as the first line in this function seems to solve this issue but this is most probably not the best solution. (I am not an expert on this...)

rwjblue:
Let the underlying subsystem deal with the `path.sep` differences, the logic here was incorrectly interchanging `\` and `/` on windows.

Fixes #48.

### DIFF
--- a/index.js
+++ b/index.js
@@ -129,8 +129,9 @@ module.exports = {
    * @param {String} relativePath - path to file within current app
    * @returns {Boolean} whether or not the file exists within the current app
    */
-  _doesFileExistIsCurrentProjectAddonModule: function(relativePath) {
-    relativePath = path.join('addon', relativePath.replace(path.join('modules', this._parentName()),''));
+  _doesFileExistIsCurrentProjectAddonModule: function(_relativePath) {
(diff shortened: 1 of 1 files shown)


nfc036:
You could also change the parameter comment from relativePath to _relativePath!?

rwjblue:
Yeah, true.  I forgot to update the docs there.  Would you mind sending a PR?

PR #51, opened 2016-10-07
cosmetics for the docs, no functional changes

nfc036:
Added an empty line and renamed the functions parameter back to relativePath. Hope that's ok. This is actually my very first pull request. And sooo sophisticated. Awesome. ;-)


rwjblue:
Hmm, I don't think I can merge with that commit in there. I think you will need to rebase out the `_parentName` commit first.

PR #52, opened 2016-10-10
return actual name not function from _parentName

nfc036:
this.parent.name is a function (if defined), this.parent.pkg.name is the expected name of the package.
Using 0.3.3 I get no coverage for any files in the app. (worked in 0.3.2)

### DIFF
--- a/index.js
+++ b/index.js
@@ -252,7 +252,7 @@ module.exports = {
     if (this.parent.isEmberCLIAddon()) {
       return this._findCoveredAddon().name;
     } else {
-      return this.parent.name;
+      return this.parent.pkg.name;
     }
   },
 
--- a/index.js
+++ b/index.js
@@ -252,7 +252,7 @@ module.exports = {
     if (this.parent.isEmberCLIAddon()) {
       return this._findCoveredAddon().name;
     } else {
-      return this.parent.pkg.name;
+      return this.parent.name();
     }
   },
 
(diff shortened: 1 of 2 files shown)


nfc036:
The failing test expects name to be the actual name (string). If I am right, name is the corresponding function [here in ember-cli](https://github.com/ember-cli/ember-cli/blob/1cae632e9724d0fc02cb91e90d83e70e78f75069/lib/models/project.js#L142). Hence, the test is incorrect. Possible?

rwjblue:
Name in an addon is a property, but in a project it is a function (which is where you linked to above).

rwjblue:
@dfreeman - Would you mind reviewing this?

index.js
```diff
@@ -252,7 +252,7 @@ module.exports = {
     if (this.parent.isEmberCLIAddon()) {
       return this._findCoveredAddon().name;
     } else {
-      return this.parent.name;
+      return this.parent.name();
```
rwjblue: Aha! This seems totally correct, sorry for the misleading comments without reading the diff carefully.

nfc036: ... sorry for the misleading comments regarding the commits... ;-)

rwjblue:
Published as 0.3.4, thanks for your help reporting and fixing this!

dfreeman:
Tests were present, just wrong (as you found) 😣  — sorry about that! Thanks for the fix, @nfc036.

rwjblue:
Thank you for reviewing @dfreeman!

PR #58, opened 2016-10-27
Add ability to pass addonName and fix paths

### LINKED ISSUE #36: Addon Support: Properly name files in `lcov.info`
Currently `lcov.info` namespaces addon code as: `SF:modules/addon-name/models/ab.js`

It should be: `SF:addon/models/ab.js` so that code coverage tools like Coveralls can properly parse the file.

I solved this using `sed s/SF:modules\\/addon-name/SF:addon/` in my build jobs, but we probably should work on properly renaming it and save everyone the trouble. If someone could point me to some lines or functions to look at for context, I'll take it on.

rwwagner90:
This allows passing `addonName` in the config and it will replace `modules/my-addon` with `addon` in the lcov.info file. This should allow proper file mapping for things like Code Climate and Coveralls. Does this approach look okay to you @rwjblue and @kategengler?

Resolves #36

### DIFF
--- a/lib/attach-middleware.js
+++ b/lib/attach-middleware.js
@@ -22,12 +22,12 @@ module.exports = function(app, options) {
 
       // If we define an addonName, strip modules/addonName and replace with 'addon'
       if (_config.addonName) {
-        for (var key of Object.keys(req.body)) {
+        Object.keys(req.body).forEach(function(key) {
           if (key.includes('modules/' + _config.addonName)) {
             req.body[key].path = req.body[key].path
               .replace(new RegExp('modules/' + _config.addonName, 'g'), 'addon');
           }
-        }
+        });
       }
 
       collector.add(req.body);
--- a/lib/attach-middleware.js
+++ b/lib/attach-middleware.js
@@ -22,10 +22,10 @@ module.exports = function(app, options) {
 
       // If we define an addonName, strip modules/addonName and replace with 'addon'
       if (_config.addonName) {
+        var regex = new RegExp('modules/' + _config.addonName, 'g');
         Object.keys(req.body).forEach(function(key) {
           if (key.includes('modules/' + _config.addonName)) {
-            req.body[key].path = req.body[key].path
-              .replace(new RegExp('modules/' + _config.addonName, 'g'), 'addon');
+            req.body[key].path = req.body[key].path.replace(regex, 'addon');
           }
         });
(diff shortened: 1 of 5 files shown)


rwwagner90:
We could probably also use this same method to remove the `tests/` output, since we don't want coverage for tests right?

offirgolan:
@rwwagner90 is there no way to get the default addon name from the addon's package.json?

lib/config.js
```diff
@@ -36,6 +36,7 @@ function config(root) {
  */
 function getDefaultConfig() {
   return {
+    addonName: null,
```
offirgolan: You can pass the root from the parent `config` method and do something like this:

`addonName: path.join(root, 'package.json').name`

lib/attach-middleware.js
```diff
@@ -20,6 +20,16 @@ module.exports = function(app, options) {
       var reporter = new Istanbul.Reporter(null, path.join(options.root, _config.coverageFolder));
       var sync = false;
 
+      // If we define an addonName, strip modules/addonName and replace with 'addon'
+      if (_config.addonName) {
+        Object.keys(req.body).forEach(function(key) {
+          if (key.includes('modules/' + _config.addonName)) {
+            req.body[key].path = req.body[key].path
+              .replace(new RegExp('modules/' + _config.addonName, 'g'), 'addon');
```
offirgolan: I dont think you need to create a new `RegExp` every time.

rwwagner90:
@offirgolan we can definitely pull from package.json, I was just going with the simplest first implementation. How would you recommend differentiating between addons and apps in that case? They would both have a `name` in package.json.

rwwagner90:
@offirgolan I updated with your suggestions, please take a look 😄

offirgolan:
@rwwagner90 lgtm 😄! Thanks for taking initiative on this.

rwwagner90:
No problem @offirgolan! I definitely want to help this addon succeed 😄

rwwagner90:
@rwjblue and @kategengler PTAL when you get a chance. It's fairly small and straight forward. Will gladly make any changes you may want to get it merged in.

arjansingh:
LGTM

rwjblue:
Thanks @rwwagner90!

rwjblue:
Published in 0.3.5

PR #60, opened 2016-10-27
Update README.md with Pretender example.

HallDJack:
The README references the need to create a pass through when using both Mirage and Pretender, but only supplies an example for Mirage. I had to figure out what was needed for Pretender and don't want the next person in the same position to have to do the same.


rwjblue:
Awesome, thank you!

PR #62, opened 2016-10-31
Factor in unused/unrequired files in to coverage %

### LINKED ISSUE #20: Does not include untested files
Is there an option to include files that are part of the `app/` folder but haven't been executed when running tests?

Not doing so results in a wrong, overstated coverage score.

kategengler:
- Solution is thanks to @wagenet
- Change node versions to test against

Should fix #20

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 ---
 language: node_js
 node_js:
-  - "0.10"
   - "0.12"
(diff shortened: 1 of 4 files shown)


rwjblue:
Lets give it a whirl!

rwjblue:
Published as 0.3.6!

PR #64, opened 2016-11-02
Avoid throwing errors while requiring files for coverage.

rwjblue:
To ensure coverage percentages are accurate we added some code to automatically require all modules in the application/addon.  If an error was detected while loading the module we annotated it with information on why it occurred, but still rethrew the error. 

In retrospect, it seems that a number of applications have dormant (and invalid) modules that exist in the app (at least during tests).  Our prior concept/assertion that rethrowing an error was the correct path turns out to itself be painful.

This changes that error into a warning (because I truly believe we should encourage folks to fix these scenarios).

Fixes https://github.com/kategengler/ember-cli-code-coverage/issues/63.

### DIFF
--- a/lib/templates/test-body-footer.html
+++ b/lib/templates/test-body-footer.html
@@ -6,8 +6,8 @@
       if (parts[0] === '{%PROJECT_NAME%}' && parts[1] !== 'tests') {
         try {
           require(file);
-        } catch(e) {
-          throw 'error requiring file for coverage: '+file;
+        } catch(error) {
+          console.warn('Error occurred while evaluating `' + file + '`: ' + error.message + '\n' + error.stack);
         }
       }
     });
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "node-test": "mocha test/**/*-test.js",
+    "browser-test": "COVERAGE=true ember test",
     "lint": "eslint config lib test tests *.js",
     "jscs": "jscs lib test",
     "test": "npm run-script jscs && npm run-script lint && npm run-script node-test"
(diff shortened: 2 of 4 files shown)


rwjblue:
@kategengler - r?

kategengler:
LGTM

kellyselden:
looks good

lib/templates/test-body-footer.html
```diff
@@ -7,7 +7,7 @@
         try {
           require(file);
         } catch(e) {
```
mike183: Seems like this param needs to be renamed from `e` to `error`. Either that or change the `console.warn()` below to use `e`.

mike183:
I was trying to debug this issue when I came across this PR! Thank you!

Apart from the minor comment above, it seems to have fixed the issue for me. I can submit a PR if for the above change if needed.

lib/templates/test-body-footer.html
rwjblue: Good catch! Updated...

rwjblue:
> I can submit a PR if for the above change if needed.

Just updated, will merge and release as soon as travis is happy...

rwjblue:
Because I found it ironic that the code coverage tool had no coverage for this issue, I went ahead and added a previously failing test...

package.json
```diff
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "node-test": "mocha test/**/*-test.js",
+    "browser-test": "COVERAGE=true ember test",
```
kategengler: The 'integration' node test runs COVERAGE=true ember test within ember-cli-code-coverage, so this might be extra work.

rwjblue: Yeah, I see it now [here](https://github.com/kategengler/ember-cli-code-coverage/blob/master/test/integration/coverage-test.js#L23-L28)

PR #68, opened 2016-11-07
Refactor add-on renaming

### LINKED ISSUE #67: Unify code to correct file names in coverage output
From #58 

There is a place that changes the displayed filenames in the instrumenter: https://github.com/kategengler/ember-cli-code-coverage/blob/master/lib/coverage-instrumenter.js#L27

Is it possible to do this massaging there?

Another potentially relevant spot: https://github.com/kategengler/ember-cli-code-coverage/blob/master/index.js#L129

rwwagner90:
Moved from attach-middleware to coverage-instrumenter, where the other path fixing occurs. Resolves #67

### DIFF
--- a/lib/coverage-instrumenter.js
+++ b/lib/coverage-instrumenter.js
@@ -1,5 +1,6 @@
 'use strict';
 
+require('string.prototype.includes');
 var existsSync = require('exists-sync');
 var extend = require('extend');
 var Filter = require('broccoli-filter');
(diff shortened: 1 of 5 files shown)


lib/coverage-instrumenter.js
```diff
@@ -24,25 +25,35 @@ function getPathForRealFile(relativePath, root, templateExtensions) {
   return null;
 }
 
-function fixPath(relativePath, name, root, templateExtensions) {
+function fixPath(relativePath, name, root, templateExtensions, isAddon) {
   // Handle apps (served from app name)
-  if (relativePath.indexOf(name) === 0) {
-    relativePath = relativePath.replace(name, 'app')
-    return getPathForRealFile(relativePath, root, templateExtensions) || relativePath
+  if (!isAddon && relativePath.includes(name)) {
```
rwjblue: Hmm, I think that `includes(name)` would return false positives wouldn't it? 

Consider running against an app named `ui`, addon files for an addon named `breadcrumb-ui` would be included...

I think we need to keep the `indexOf(name) === 0` or swap to `startsWith(name)`.

lib/coverage-instrumenter.js
```diff
@@ -24,25 +25,35 @@ function getPathForRealFile(relativePath, root, templateExtensions) {
...
+    relativePath = relativePath.replace(name, 'app');
+    return getPathForRealFile(relativePath, root, templateExtensions) || relativePath;
   }
 
   // Handle addons (served from dummy app)
-  if (relativePath.indexOf('dummy') === 0) {
-    relativePath = relativePath.replace('dummy', 'app')
-    var dummyPath = path.join('tests', 'dummy', relativePath)
+  if (relativePath.includes('dummy')) {
```
rwjblue: Like above, it should be `indexOf(name) === 0` or `.startsWith(name)`.  Also, should change the conditional to only do this when `isAddon`.

rwjblue: After reading the below section, it might be good to just have a top level `if (isAddon) { } else { }`, so we can keep it super clear which things apply to each thing and don't have to repeat the guard check for each "concept".

Something like:

```js
if (isAddon) {
  // Handle addon's dummy app
  if (relativePath.indexOf('dummy') === 0) {
    relativePath = relativePath.replace('dummy', 'app');
    var dummyPath = path.join('tests', 'dummy', relativePath);
    return (
      getPathForRealFile(dummyPath, root, templateExtensions) ||
      getPathForRealFile(relativePath, root, templateExtensions) ||
      relativePath
    );
  }

  // Handle addons renaming of modules/my-addon
  if (relativePath.indexOf('modules/' + name) === 0) {
    var regex = new RegExp('modules/' + name, 'g');
    relativePath = relativePath.replace(regex, 'addon');
    return (
      getPathForRealFile(relativePath, root, templateExtensions) ||
      relativePath
    );
  }

  return relativePath;
} else {
  relativePath = relativePath.replace(name, 'app');
  return getPathForRealFile(relativePath, root, templateExtensions) || relativePath;
}
```

lib/coverage-instrumenter.js
```diff
@@ -24,25 +25,35 @@ function getPathForRealFile(relativePath, root, templateExtensions) {
...
       getPathForRealFile(relativePath, root, templateExtensions) ||
       relativePath
-    )
+    );
   }
 
-  return relativePath
+  // Handle addons renaming of modules/my-addon
+  if (isAddon && relativePath.includes('modules/' + name)) {
+    var regex = new RegExp('modules/' + name, 'g');
```
rwjblue: Should be anchored to the beginning

lib/coverage-instrumenter.js
rwwagner90: Good call, will fix

PR #79, opened 2016-11-23
Add include option to add additional coverage to the report

jesseditson:
This PR adds an `include` option to the config hash that will allow users to include additional coverage to the output.

My current setup is as follows:

- I have application server code that lives in `server`
- I have a suite of tests for app server code that lives in `test`
- I run server tests on rebuild using an in-repo addon
- The server tests output to `coverage-server`, which I'd like to include in the coverage report

This option allows me to specify the `coverage-server/coverage-final.json` (generated via `nyc`), and get accurate coverage on my entire app, instead of having 2 different coverage reports.


jesseditson:
Relates to: https://github.com/ember-cli/rfcs/issues/81 - with that RFC and this PR, tests & coverage would work across both the `app` and the `server` code

rwwagner90:
This seems okay to me. @rwjblue @kategengler @kellyselden thoughts? I would definitely like to see some tests added here, but I am not opposed to the idea of this feature.

PR #83, opened 2016-12-02
Add coverage-merge command and parallel configuration

jdenly:
Allows use of parallel testing with ember-exam or similar.

Additional configuration flag `parallel`, when set to true will output coverage into multiple directories, with a random suffix. The `coverage-merge` command can then be used after the build to merge them together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ and then:
 
 `cross-env COVERAGE=true ember test`
 
+When running with `parallel` set to true, the final reports can be merged by using `ember coverage-merge`. The final merged output will be stored in the `coverageFolder`.
+
 ## Configuration
 
 Configuration is optional. It should be put in a file at `config/coverage.js`. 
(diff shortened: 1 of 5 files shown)


gustaff-weldon:
@jdenly that sounds great. One question. Since we do not use any external coverage service that we can upload file to, how

jdenly:
@gustaff-weldon I'm not seeing coverage reports being displayed on the console at all. How do you run your tests? What type of reporters do you have configured?

This just changes the coverage writing mechanism to write to multiple directories, coverageFolder suffixed with a random string, for example coverage_abcdef12 and coverage_efef1212. Once you're finished, you can merge them into the coverageFolder using ember coverage-merge.

`ember exam --split=4 --parallel && ember coverage-merge`

You need to manually remove all coverage folders between runs - this is sufficient for our use case of running on a CI server as it cleans up between runs. Thinking about it it might be better to delete the coverage_* directories after merge.

pleszkowicz:
@jdenly you can add/change reporters in the [config](https://github.com/kategengler/ember-cli-code-coverage#options)]
Available options: https://github.com/gotwarlost/istanbul/tree/master/lib/report
If you need to display on the console, please set `reporters: ['text']`.

kategengler:
@jdenly Thanks for this and sorry for taking so long to respond to this. I was wondering if you'd be willing to add a bit of a smoke test for this, similar to what we have https://github.com/kategengler/ember-cli-code-coverage/blob/master/test/integration/coverage-test.js#L23 but checking for the results of the merged report?

I won't block merging this PR on it (I'm about to do so) but would really appreciate the test.

jdenly:
@kategengler No worries, I've added a test under #92.

PR #87, opened 2016-12-29
Use synchronous control flow when making synchronous requests

DingoEatingFuzz:
The onload event for XHRs does not fire when the XHR is synchronous.
This was causing the callback to never be called and the test suite
to hang indefinitely.

See https://github.com/kategengler/ember-cli-code-coverage/pull/82 for the discussion that led up to this.


PratheepV:
It worked for me. Will it be merged ?

PratheepV:
Cool.

Issue #75, opened 2016-11-07
Coverage failing with version 0.3.6 and up


rox163:
I seem to have an issue with v 0.3.6+ of ember-cli-code-coverage. 
Running `COVERAGE=true ember test` kept failing after the last test was run ie when coverage would normally start. Its not repo specific. Doesnt matter what the tests are, this error always shows up at the end
```
not ok 29 Chrome 54.0 - FrostNavigationService "after each" hook
    ---
        message: >
            Uncaught TypeError: Cannot convert undefined or null to object (http://localhost:7357/5392/tests/index.html?hidepassed:43)
        stack: >
            Error: Uncaught TypeError: Cannot convert undefined or null to object (index.html?hidepassed:43)
                at global.onerror (http://localhost:7357/assets/test-support.js:8732:10)
        Log: |

```
I have pin pointed this to version 0.3.6 of ember-cli-code-coverage. Up to 0.3.5 everything works great. If you pin to v 0.3.5, coverage is reported properly.  Could be related to this PR https://github.com/kategengler/ember-cli-code-coverage/pull/62/files
@rwjblue @rwwagner90

rwjblue:
Can you tell me if `Object.keys(require.entries)` causes the same error?

If it does, this is likely the same (or related to the same root cause) as https://github.com/kategengler/ember-cli-code-coverage/issues/66.

rox163:
@rwjblue Not sure. What would you like me to do to test that `Object.keys` line?

rwjblue:
Run `ember test --server`, after your tests run in your browser open the console, type `Object.keys(require.entries)`. Is there an error?

rox163:
@rwjblue Yes there is an error.
![screen shot 2016-11-08 at 10 07 12 am](https://cloud.githubusercontent.com/assets/313318/20104241/3babba8a-a59b-11e6-8afd-81e9afd54520.png)

rox163:
Any suggestions on how to debug/work around this? As far as I can tell there is no custom `require` in my project directly but not sure about the dependent addons. We dont use yadda as was mentioned as the culprit in the other issue. 
I can point you to a project that has this failure when the coverage version is set to 0.3.6 - https://github.com/ciena-frost/ember-frost-navigation/blob/master/package.json.

rox163:
Related to https://github.com/kategengler/ember-cli-code-coverage/issues/66 as in my case it seems to be a conflict with mocha's require.

[...]

PR #97, opened 2017-01-31
fix require.entries is undefined

a15n:
fixes the issue discussed here https://github.com/kategengler/ember-cli-code-coverage/issues/75


rwwagner90:
According to #75 I think we do not want to merge this in, as it should explode if `require` is overwritten by other addons. Going to close for now, and if people continue to have issues, happy to explore other solutions.

PR #99, opened 2017-02-18
Update babel-core

mike183:
Resolves a deprecation warning being issued due to `babel-core` using an outdated version of `minimatch`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "ember-addon"
   ],
   "dependencies": {
-    "babel-core": "^5.8.38",
+    "babel-core": "^6.23.1",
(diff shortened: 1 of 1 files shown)


mike183:
Tests seem to be passing on local so I'm not sure if they're related to this change or not?

Does [this line](https://github.com/kategengler/ember-cli-code-coverage/blob/master/lib/coverage-instrumenter.js#L70) need to be changed in some way?

mike183:
As per a quick conversation with @kategengler (albeit a while ago), this will most likely need some more in-depth work.

I don't currently have the time or babel experience to put in the necessary work, therefore I'm closing this PR.

PR #100, opened 2017-02-22
Fixes #42 use configPath instead of hard-coded config

### LINKED ISSUE #42: Use `configPath` instead of hard-coded "config"
Ember CLI supports a custom path for app's configuration via `configPath` variable in `package.json`. `Project` model has [`configPath`](https://github.com/ember-cli/ember-cli/blob/2bce175ce188ae0f47a55fa23d6affe2d768ddfb/lib/models/project.js#L177) method to aid.

I think it could be as easy as replacing `config` to `configPath` function call over [here](https://github.com/kategengler/ember-cli-code-coverage/blob/8257142489a7e5ac0847db3787205971370442c3/lib/config.js#L22).

eddie-ruva:
On this PR I'm trying to tackle the issue brought up in #42 by using `configPath` instead of the hard coded config path value. 

According to the [ember-cli docs](https://ember-cli.com/api/classes/Project.html#method_configPath) `configPath` is private but I couldn't figure out a better/public way to determine this value. If anybody knows of a better way to make this change I'll be happy to integrate the feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ When running with `parallel` set to true, the final reports can be merged by usi
 
 ## Configuration
 
-Configuration is optional. It should be put in a file at `config/coverage.js`. 
+Configuration is optional. It should be put in a file at `config/coverage.js` (`configPath` configuration in package.json is honored)
 
 #### Options
 
--- a/index.js
+++ b/index.js
@@ -77,7 +77,7 @@ module.exports = {
 
   testemMiddleware: function(app) {
     if (!this._isCoverageEnabled()) { return; }
-    attachMiddleware(app, { root: this.project.root, ui: this.ui });
+    attachMiddleware(app, { configPath: this.project.configPath(), root: this.project.root });
   },
 
   // Custom Methods
@@ -204,7 +204,7 @@ module.exports = {
    * @returns {Configuration} project configuration
    */
   _getConfig: function() {
-    return config(this.project.root);
+    return config(this.project.configPath());
   },
 
   /**
--- a/lib/attach-middleware.js
+++ b/lib/attach-middleware.js
@@ -17,7 +17,7 @@ module.exports = function(app, options) {
     bodyParser.json({ limit: '50mb' }),
     function(req, res) {
       var collector = new Istanbul.Collector();
-      var _config = config(options.root);
+      var _config = config(options.configPath);
 
       if (_config.parallel) {
         _config.coverageFolder = _config.coverageFolder + '_' + crypto.randomBytes(4).toString('hex');
--- a/lib/config.js
+++ b/lib/config.js
@@ -15,11 +15,12 @@ var path = require('path');
 /**
  * Get configuration for a project, falling back to default configuration if
  * project does not provide a configuration of its own
- * @param {String} root - root path of project
+ * @param {String} configPath - The path for the configuration of the project
  * @returns {Configuration} configuration to use for project
  */
-function config(root) {
-  var configFile = path.join(root, 'config', 'coverage.js');
+function config(configPath) {
+  var configDirName = path.dirname(configPath);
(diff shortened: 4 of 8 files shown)


README.md
```diff
@@ -33,7 +33,7 @@ When running with `parallel` set to true, the final reports can be merged by usi
 
 ## Configuration
 
-Configuration is optional. It should be put in a file at `config/coverage.js`. 
+Configuration is optional. It should be put in a file at named `coverage.js` inside your ember app config folder (defaults to `config/`).
```
rwjblue: I'd prefer to keep this README line unchanged. 99% of our users will be using `config/coverage.js`, and I think rewording like jisnjust makes it hard to grok for the vast majority use case. I think a parenthetical aside or a second sentence mentioning that we honor `configPath` is enough...

index.js
```diff
@@ -77,7 +77,10 @@ module.exports = {
 
   testemMiddleware: function(app) {
     if (!this._isCoverageEnabled()) { return; }
-    attachMiddleware(app, { root: this.project.root, ui: this.ui });
+    attachMiddleware(app, {
+      root: this.project.root,
+      configDirName: path.dirname(this.project.configPath())
```
rwjblue: Why are we no longer passing in `ui`?

Can you put the config path invocation above here as a variable declaration (then just pass it through here)?

rwjblue: On second thought, I think it would be better to simply pass the project down into the middleware and have our config util function accept the project and avoid the path munging throughout the rest of the system.

README.md
eddie-ruva: Thanks for taking a look at this! I've reworded, let me know if the new message is more inline with what you have in mind.

index.js
eddie-ruva: 1. I removed `ui` because it was not being reference from the middleware so the parameter was unnecessary. 

2. I changed the config code to now take in the full `project` and construct the path inside.

rwjblue:
@kategengler - Any objections?

lib/config.js
```diff
@@ -15,11 +15,12 @@ var path = require('path');
 /**
  * Get configuration for a project, falling back to default configuration if
  * project does not provide a configuration of its own
- * @param {String} root - root path of project
+ * @param {Object} project - The Project model tied to the package.json
  * @returns {Configuration} configuration to use for project
  */
-function config(root) {
-  var configFile = path.join(root, 'config', 'coverage.js');
+function config(project) {
```
kategengler: I'd prefer to continue to pass in the config path, rather than pass around the `project` (I'd like to keep usage of ember-cli api contained to index.js as much as possible)

lib/attach-middleware.js
```diff
@@ -12,12 +12,12 @@ function logError(err, req, res, next) {
   next(err);
 }
 
-module.exports = function(app, options) {
+module.exports = function(app, project) {
```
kategengler: I think this should still take options and get a configPath and root as properties

lib/config.js
eddie-ruva: I think this makes sense because I could still keep the logic to construct the path to `coverage.js` inside `config` like @rwjblue suggested but also not take in the whole `project` object. 

I can make this change if there are no objections.

rwjblue: 👍 - Thanks for iterating on this with us

eddie-ruva: @kategengler @rwjblue No problem! I've updated with the feedback received. Let me know if you notice anything else and thanks for taking the time to review this

kategengler:
Looks great, thanks for making those changes @eddie-ruva

alias-mac:
@kategengler, any estimate when will this be released in the next version?

rwjblue:
Released now as v0.3.12.

Issue #85, opened 2016-12-06
config/coverage options for excludes don't seem to work


jbelmont:
I have this version of ember-cli-code-coverage in my package.json "ember-cli-code-coverage": "0.3.8". Also I have a coverage file in this root path in my ember app. 

config/coverage.js
```
module.exports = {
     // This is default value
    coverageEnvVar: 'COVERAGE',
     // More reporters found here https://github.com/gotwarlost/istanbul/tree/master/lib/report
    reporters: ['lcov', 'json', 'json-summary', 'text', 'text-summary'],
    // Defaults to ['*/mirage/**/*']
    excludes: [
        "/blueprints",
        "/config",
        "/public", 
        "/tmp", 
        "/tests/dummy/**",
        "/vendor"
    ],
    // Defaults to coverage. A folder relative to the root of your project to store coverage results.
    // Set to true or false if you are using ESNext features.
    useBabelInstrumenter: true
}
```

I have the following folder structure
➜  ember-webrtc-devices git:(master) tree -d -L 3 -I 'node_modules|bower_components'
.
├── addon
│   ├── components
│   │   └── device-selection
│   └── mixins
├── app
│   ├── components
│   │   └── device-selection
│   ├── mixins
│   └── translations
│       └── webrtc-devices
├── blueprints
│   └── ember-webrtc-devices
├── config
├── coverage
│   └── lcov-report
│       ├── addon
│       ├── app
│       └── tests
├── public
│   └── assets
│       └── sounds
├── scripts
├── tests
│   ├── dummy
│   │   ├── app
│   │   ├── config
│   │   └── public
│   ├── helpers
│   ├── integration
│   │   └── components
│   └── unit
│       └── mixins
├── tmp
├── typings
│   └── ember
└── vendor

36 directories

When I run `npm run coverage` which runs   "coverage": "COVERAGE=true ember test"

I got the following code coverage report.

------------------------------------|----------|----------|----------|----------|----------------|
File                                |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
------------------------------------|----------|----------|----------|----------|----------------|
 addon/components/device-selection/ |     62.5 |    53.85 |     62.5 |    57.14 |                |
  component.js                      |     62.5 |    53.85 |     62.5 |    57.14 |... ,97,115,126 |
 addon/mixins/                      |    29.82 |    19.74 |    33.33 |    25.23 |                |
[...]

Issue #76, opened 2016-11-08
RFC: API to cover both adding and removing files from coverage


kellyselden:
Now that we are requiring all files, issues like #65, #66, and #75 are popping up, and the problem may not be immediately clear. We also have files we want to stop covering in #71. I propose we unify any API to resolve these issues, or at least make them compatible. Maybe both a `shouldInclude` and a `shouldExclude` hook. That's about as far as I got in the thought process.

rwwagner90:
@kellyselden there is an `exclude` array, but I do not think it reliably works. We could have a sane `exclude` default, and then the user could override it to be blank if they wanted to include everything or add to it to exclude more, etc. Thoughts?

Issue #71, opened 2016-11-07
Stop covering `tests/dummy/app` files.


rwjblue:
It seems pretty odd that we would want to consider `tests/dummy/app` files as part of code coverage (since they are actually part of the test harness itself).

Does anyone have objections?

/cc @sglanzer @kategengler @rwwagner90 @sandersky @kellyselden

rwjblue:
FWIW, this will require the changes discussed in https://github.com/kategengler/ember-cli-code-coverage/issues/69 to implement (though not particularly hard).

kellyselden:
If you have a complicated acceptance test setup with a large dummy app, and test coverage tells you that some of that dummy app is not being run, would that be something worth preserving?

rwjblue:
> If you have a complicated acceptance test setup with a large dummy app, and test coverage tells you that some of that dummy app is not being run, would that be something worth preserving?

Ya, I can definitely see this argument.

To me it comes down to what the point of "code coverage" is. IMO, the goal is to ensure that the code we are shipping to consumers (`addon/` and `app/` for addons, and `app/` for apps) is sufficiently covered. 

Also, there are a large number of addons that primarily use `tests/dummy/app` for their demo / marketing sites. Much of that code is presentational and may or may not be tested...

kellyselden:
Would it be too complex to not cover "dummy" by default, but allow a config option to turn it on?

rwwagner90:
@kellyselden I approve of disabling by default and allowing an option to turn it on. Would not be hard to do at all.

rwjblue:
Unsure, but we should investigate when we implement.  I agree having finer grained control of "should cover" or "should not cover" would be great....

sglanzer:
I've used an `excludes` filter that was configurable for something like this before, I agree with not including the dummy by default for the same reasons @rwjblue already listed

kellyselden:
I also agree that not covering `dummy` by default is a safe choice.

rwwagner90:
If everyone agrees, I will remove dummy by default and add a config flag to enable dummy coverage.

rwjblue:
I agree that we should remove it from being included by default, but I would prefer to not add a flag specifically for `includeDummy` to config.

I would like to just go whole hog and add a `shouldIncludeFile` hook to config.

Might look like:

```js
  module.exports = {
    shouldIncludeFile: function(project, relativePath, includedByDefault) {
      return; // do the default behavior

      return false || true;
    }
[...]

Issue #41, opened 2016-09-18
Excludes not working as expected with addon


alexlafroscia:
Having now set up coverage for `ember-ajax`, I'm trying to get it working right.

The follow are the files that are having coverage calculated:

<img width="302" alt="screen shot 2016-09-17 at 6 31 24 pm" src="https://cloud.githubusercontent.com/assets/1645881/18612089/022466b8-7d05-11e6-819d-a9a358849eda.png">

However, I've tried to ignore the whole `tests/dummy` directory with the configuration [here](https://github.com/ember-cli/ember-ajax/blob/69df203c3b396142a4462186285c198d6df44c9a/config/coverage.js).  It seems like the files are not actually being ignored.

srowhani:
I found that `/dummy/(.*)/` or something similar does the trick.

rwwagner90:
@alexlafroscia still having this issue?

konradjurk:
@alexlafroscia I ran into the same issue and were able to resolve it this way:

1. First change the output directory in your coverage config temporarily, just to make sure it is actually hit (`coverage.js` has to live under `configPath` from package.json).
2. Add a function to the `excludes` array that prints out the first positional argument which is the file path. This function is called for every file that the addon finds.

      Most paths are unexpectedly different to the ones you see in the IDE (e.g. `app` file tree is not present, probably because the dummy app is built so `dummy/app/router.js` becomes `dummy/router.js`).

    From here you can adjust your excluded files array so it is working as expected :)

rwwagner90:
Thanks for the feedback @konradjurk! I think with this workaround, everything should be working as expected. Would you be interested in submitting a PR adding something about setup for addons to the README?

rwwagner90:
I believe this should be resolved in the latest version, 1.0.0-beta.2. If it is not, please feel free to reopen!

PR #101, opened 2017-03-08
WIP: Exclude dummy files

sfbwalder:
Relates to issues #41, #71, and #85.  It looks like there is a desire to not include the dummy app files in an add-on.  We also need them excluded.

The desired solution is documented in #76, but there there hasn't been any movement there.  Can a short term solution be implemented to always exclude them?  Or is there a requirement to include them?

The [logic](https://github.com/kategengler/ember-cli-code-coverage/blob/master/index.js#L193) is always including modules in the dummy app rather than excluding them.  This change always excludes them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -189,14 +189,20 @@ module.exports = {
       relativePath = relativePath.replace('dummy/', '');
     }
 
-    var fileExists = (
-      this._doesFileExistInDummyApp(relativePath) ||
-      this._doesFileExistInCurrentProjectApp(relativePath) ||
-      this._doesFileExistInCurrentProjectAddon(relativePath) ||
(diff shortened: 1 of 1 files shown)


kellyselden:
Should this be opt-in? For example, you've made a change that unexpectedly causes a section of your dummy app to not run in an acceptance test. Or do you rely on your test failing at that point?

sfbwalder:
Is this the direction we want this to go to for #76 and as suggested here: https://github.com/kategengler/ember-cli-code-coverage/issues/71#issuecomment-258931661?  It allows an opt-in for test/dummy/ files.

I don't have the tests complete, but if this is a good direction I will try to finish finish this up.

@kellyselden @kategengler @rwjblue

Issue #108, opened 2017-04-19
Parsing async/await crashes


sukima:
This appears related to a old version of istanbul: https://github.com/gotwarlost/istanbul/issues/733

```console
Failed to parse file: app/components/my-component.js
Unable to cover: app/components/my-component.js . Try setting useBabelInstrumenter to true.
 Error: Line 66: Unexpected identifier
  at constructError (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2407:21)
  at createError (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2426:17)
  at unexpectedTokenError (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2500:13)
  at throwUnexpectedToken (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2505:15)
  at expect (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2523:13)
  at expectCommaSeparator (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2547:13)
  at parseObjectInitializer (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:3053:17)
  at inheritCoverGrammar (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2681:18)
  at parsePrimaryExpression (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:3247:20)
  at inheritCoverGrammar (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2681:18)
  at parseLeftHandSideExpressionAllowCall (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:3414:20)
  at inheritCoverGrammar (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:2681:18)
  at parsePostfixExpression (/Users/suki/source/SeeClickFix/partners/node_modules/istanbul/node_modules/esprima/esprima.js:3490:16)
[...]

PR #109, opened 2017-04-21
Improve ES7 error message

### LINKED ISSUE #106: Cannot parse async/await

sukima:
This messaging was confusing to me so I took a stab at clarifying the intent.

Fixes #106 and #108


rwjblue:
Awesome, thank you!

PR #119, opened 2017-06-01
Add mocha support and fix pretender bug

mattmarcum:
This should address https://github.com/kategengler/ember-cli-code-coverage/issues/2 and also fixes a bug I was seeing with the pretender passthrough/fake request not having a `response` property.


apellerano-pw:
Thanks for working on this, I was just wondering why lcov wasn't regenerating when our mocha tests re-ran on the `/tests` endpoint of `ember serve`!

mattmarcum:
I found a different pretender bug....I finally figured out that some of my unit tests weren't tearing mirage down properly by calling `window.server.shutdown()`.  Even though I had the passthrough enabled in my mirage configs, code coverage would never show up in the browser.  It was failing silently because the `request.onload` function in the `test-footer` script was never getting called.  I think this is a pretender bug caused by a new pretender instance being created multiple times and never being torn down.

rwjblue:
FWIW, pretender throws now if you create a second instance without shutting down the last

mattmarcum:
Is there anything I need to change about this PR?  I noticed that it's failing travis ci builds for node v 0.12 - because ember-cli-htmlbars-inline-precompile isn't transpiled: 

```
  `ember test`
/home/travis/build/kategengler/ember-cli-code-coverage/node_modules/ember-cli-htmlbars-inline-precompile/index.js:15
  init() {
      ^
Unexpected token (
```

paulcwatts:
For what it's worth, https://github.com/kategengler/ember-cli-code-coverage/pull/115 drops support for Node 0.12.

mattmarcum:
Anybody got any suggestions for the failing tests?  When I run them locally with `ember test` on node v6 they all pass.  There's an error in the travis console log that says and address is in use and I'm thinking that's the root cause.

paulcwatts:
@mattmarcum Are you running the tests through "npm test", instead of "ember test"? In those specific test cases, "ember test" is run through the node tests.

Those EADDRINUSE errors usually occur when a preceeding test fails, and the "ember test" child doesn't exit or hangs, so the next test runs "ember test" and the port is in use.

mattmarcum:
ok I ran them with `npm test` and they all (47) pass locally.  I'll look some more when I get home.

paulcwatts:
The tests are also fairly flaky, as I've noticed. They can fail intermittently on Travis.

mattmarcum:
Yeah, I squashed my commits and pushed up and now the node v4 tests fail and the node v6 tests pass - the exact opposite of what was happening before - no code changes.

rwwagner90:
@mattmarcum seems the tests are a bit flaky, but I restarted them and they passed 🎉 . Going to go ahead and merge this in.

PR #126, opened 2017-06-30
Suggest using `posttest` hook for `ember coverage-merge`

luketheobscure:
It might be helpful to suggest how/where to run `coverage-merge` for people looking to set this up in a CI environment.

The only other thing I might add is that you need to run `npm test` and have the appropriate configuration there for `posttest` to trigger, but I'm not sure if that's overkill since that should be common knowledge.


kellyselden:
LGTM

PR #127, opened 2017-06-30
Resolve addon file paths correctly in CLI >= 2.12

dfreeman:
Starting in (I believe) ember-cli@2.12, addon files stopped being nested under `modules/`, which means the path correction logic here wasn't working properly for them.

If you have thoughts on the best way to test this change, I'd welcome them 🙂

### DIFF
--- a/lib/coverage-instrumenter.js
+++ b/lib/coverage-instrumenter.js
@@ -38,10 +38,10 @@ function fixPath(relativePath, name, root, templateExtensions, isAddon) {
       );
     }
 
-    // Handle addons renaming of modules/my-addon
-    if (relativePath.startsWith('modules/' + name)) {
-      var regex = new RegExp('^modules/' + name);
-      relativePath = relativePath.replace(regex, 'addon');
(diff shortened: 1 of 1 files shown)


rwwagner90:
@dfreeman could we use ember-try and run coverage reports and check that things are renamed as expected? Aside from that, we could probably just do a quick test that passing the different file names yields the expected results here. Would definitely like at least a test that ensures the regex works as intended, if not some sort of full test.

dfreeman:
Unfortunately ember-try doesn't support swapping out versions of the CLI itself, but I can definitely add unit test coverage at least — I think I missed that this repo had a node-land test suite when I originally wrote this.

rwwagner90:
@dfreeman yeah, any sort of node tests to just lightly check this you can add would be fantastic, and then I will gladly merge 😄

rwwagner90:
With approval from @rwjblue and since he told me he is planning a large refactor of this repo anyway, I will merge without tests. Thanks for the PR @dfreeman!

PR #138, opened 2017-09-24
Update ember and bump deps

rwwagner90:


rwwagner90:
@kellyselden @rwjblue any ideas as to why the tests are failing from bumping these deps?

PR #141, opened 2017-10-02
[WIP] Refactor to use babel-plugin-istanbul

rwjblue:
This has some negative effects still:

* Does not re-write the paths to match "real" on disk paths (I left an inline TODO and a stub function to make this "easy" to add back in)
* Instruments test files (need to add them to an excludes list)

Even with these negative side-effects, it has massive upside:

* Massively less overall code to maintain
* Does not require us to parse babel config (and therefore avoids issues
  around parallelism in broccoli-babel-transpiler)
* Significantly faster when used (e.g. no longer has to double parse and
  process files)
* Loads istanbul configuration from `.istanbul.yml` file (very little documentation about this, but we are going through the default istanbul code path that loads config)

TODO:

- [ ] Fix issue with source not being include in reports
- [ ] Fix paths so they match with real paths on disk (generally needs to be done in the middleware section)


theseyi:
Will this get merged in the near term? @rwjblue or are the issue mentioned above blockers?

lib/attach-middleware.js
```diff
@@ -1,46 +1,52 @@
...
 
-      if (_config.reporters.indexOf('json-summary') === -1) {
-        _config.reporters.push('json-summary');
-      }
+      Object.keys(coverage).forEach(filename =>
+        map.addFileCoverage(fixFilePaths(coverage[filename]))
+      );
+
+      reporter.addAll(config.reporters);
+      reporter.write(map);
```
kategengler: Will this use the above `config.coverageFolder`?

srini88:
Great to see this PR. This is kind of a blocker for us. Waiting for this to be merged :-)

rwjblue:
FWIW, there are still some issues with this approach to be ironed out (mentioned some of them in the TODO's in the description), but I don't have the bandwidth to continue pushing it forward. 

Definitely would love someone to take a stab at test driving things and PR'ing fixes back to this branch...

igbopie:
Hi! I have a question. Ember has the 'knowledge' how to build our code, isn't it possible to leverage that 'knowledge' and just use source maps to get the real source code?.

Thanks!

rwwagner90:
Thanks to the great work by @adamjmcgrath this is ready and working! There may still be some things we need to add support for, but I am going to merge this in. I will release a beta version, and it would be great if people could help test it out 😃

PR #144, opened 2017-11-14
Typescript initial support

igbopie:
Hi!

I would like to add support to TS. My initial idea was to transpile TS to JS and pass that JS to BABEL using Babel-instrumenter. Then I realize that ember already knows how to build the code, why do we need to specify to coverage how to build the code? So I changed the addon hook to `postBuild`, that way I get the compiled src and the srcMaps. I read the sourceMaps and I use that to display the code. 

This is a first approach. What do you think?

Thanks!


igbopie:
Yes, not all test are passing. The reason is that inline source map is not generated on those cases. I have to figure out how, or may don't even read src map in that case.

rwjblue:
In general, we should avoid landing large changes to master until after https://github.com/kategengler/ember-cli-code-coverage/pull/141 has landed...

igbopie:
Sounds good!

rwwagner90:
@igbopie there was a pretty significant refactor recently. Would you mind reimplementing with that as the base?

rwwagner90:
@igbopie I'm not sure what is going on here, but I think TS should be supported from @dfreeman's PR we just merged https://github.com/kategengler/ember-cli-code-coverage/pull/164

igbopie:
I will take a look! Thank you

PR #145, opened 2017-11-19
[WIP] Refactor to use babel-plugin-istanbul cont.

adamjmcgrath:
Per https://github.com/kategengler/ember-cli-code-coverage/pull/141#issue-262133776

> This has some negative effects still:
> - Does not re-write the paths to match "real" on disk paths (I left an inline TODO and a stub function to make this "easy" to add back in)
> - Instruments test files (need to add them to an excludes list)

- only instrument the contents of the `app` and `addon` folder using the `includes` option for `babel-plugin-istanbul`
- Keeps a map to lookup the correct filenames by ember module name so that istanbul can find the real paths on disk
- I also removed the parallel config and merge commands. We can create a single coverage map and keep populating it with the parallel test runs. The report for each parallel test run will overwrite the previous one, but the last report will win - which will be the complete report.

TODO:
I can add some more index unit tests and update the docs if you think this is the right direction. I'd also need to remove the test fixtures I've added from the shipped addon.


rwwagner90:
Thanks for this @adamjmcgrath! Somehow I just noticed this. Will try to take a look at this soon.

rwwagner90:
@adamjmcgrath I spoke with @rwjblue and he thinks this looks pretty good, so definitely some more tests and some docs would be great 😄 . Please let me know what you need help with and I'm happy to jump in.

adamjmcgrath:
Sounds good thanks, will take a look at this in the week

adamjmcgrath:
Have made some updates:

* Taken out my suggested changes for the `parallel` option (will raise another PR for that)
* Remove test fixtures when this addon is being consumed
* Added some unit tests for the new instrumentation logic
* Couple of small updates to the docs

rwwagner90:
Thanks @adamjmcgrath! @rwjblue do you have time to do a quick review on this?

rwwagner90:
@adamjmcgrath should we add something to the docs about configuring with `.istanbul.yml`? I believe that is the path forward, versus maintaining a separate config here, right?

adamjmcgrath:
> @adamjmcgrath should we add something to the docs about configuring with .istanbul.yml? I believe that is the path forward, versus maintaining a separate config here, right?

@rwwagner90 Yes, although I haven't made any changes to support this - you would currently still need to declare `excludes`, `reporters` and `coverageFolder` in the addon's config file. 

`excludes` config could just be removed in favour of using `.istanbul.yml`. But in order to remove `reporters` and `coverageFolder` from the addon config there would need to be a change to how `parallel` works (which I intend to propose in another PR).

rwwagner90:
I ran this locally against some of my apps and addons and it seems to more or less work. This is a huge improvement from the previous version, as before it did not work at all with newer ES features like async/await. Going to go ahead and merge. Thanks so much for the PR @adamjmcgrath! If you are interested in helping out more, I could use some help going through the open issues and seeing what has been fixed by this PR and what we still need to implement.

adamjmcgrath:
> If you are interested in helping out more, I could use some help going through the open issues and seeing what has been fixed by this PR and what we still need to implement.

@rwwagner90 sure - I'm happy to help out. Will have a look through the open issues this week.

Issue #151, opened 2018-01-22
1.0.0-beta.0 not instrumenting application files


wireframe:
Testing out latest release 1.0.0-beta.0 and the test coverage report is reporting that only 32 lines exist in my app -- which is just a weee bit off :)

I'm using a vanilla install of ember-cli-code-coverage with no additional configuration ATM, and I'm happy to supply additional information and test various config to help identify the issue.

adamjmcgrath:
Are you a Windows user? I suspect there may be an issue for Windows users with generating the module paths in https://github.com/kategengler/ember-cli-code-coverage/blob/master/index.js#L189

wireframe:
Nope.  macOS user, but we see the exact same coverage report when running on our CI server (CircleCI).

kategengler:
@wireframe Any more info here? Can you reproduce in a repo you can share?

wireframe:
I wish I could share our app, but unfortunately, I can't. 😢 

I'm guessing this isn't an issue with a vanilla new ember app, so it likely has to do with conflicts in our env.  if there's any info I can provide to help troubleshoot, I'm happy to do so (ex: package.json).  For reference, we were long time users of ember-cli-blanket which worked great and we transitioned to ember-cli-code-coverage for a while (needed to remove it due to some random segfaults in CircleCI).  So we're coming back to the plugin with the latest version to test out if the issues still persist.

kategengler:
@wireframe I imagined you wouldn't be able to, I was wondering if you could try to recreate it in a vanilla app by adding things that simulate your app until it doesn't work?

rwwagner90:
@wireframe do you happen to have in-repo addons in your app?

wireframe:
Correct!  We do have two or three small inrepo addons in our app. We also use pods to organize our app Incase that info is helpful.

rwwagner90:
@wireframe can you please try 1.0.0-beta.2? I believe it should fix your issues.

wireframe:
Awesome!  1.0.0-beta.2 is now reporting 88% test coverage for app which seems *much* more accurate!  Thanks!

PR #158, opened 2018-02-02
Start in-repo addon tests

rwwagner90:
@adamjmcgrath I have been working on using ember-cli-addon-tests to spin up apps to test with and there are some issues.

Most of the coverage checks, which were 50% before, are now 83.33%. This may still be correct, since we are running in a "real" app now, but wanted to confirm with you.

The real issue, is when I start working with in-repo addons. Once I add an in-repo addon, I start getting just a flat 0% for all files in the app, but I do get % for things in the in-repo addon.

On a related note, if I have `app/utils/foo.js` and something with the same name in my in-repo addon `lib/my-addon/app/foo.js` it only reports on the latter, and not the former.

Hopefully this is making sense to you, not sure if I am being clear in my wording. I'm opening this PR so you can see my work towards the in-repo addon tests, and if you have some time to help debug, I would very much appreciate it! 😃


adamjmcgrath:
> Most of the coverage checks, which were 50% before, are now 83.33%. This may still be correct, since we are running in a "real" app now, but wanted to confirm with you.

Yeah, I would expect this because in addition to the fixtures, you'll also have `resolver.js`, `router.js`, `app.js` etc. 

> The real issue, is when I start working with in-repo addons. Once I add an in-repo addon, I start getting just a flat 0% for all files in the app, but I do get % for things in the in-repo addon.

I have a hunch as to what's happening here, will take a look

adamjmcgrath:
> The real issue, is when I start working with in-repo addons. Once I add an in-repo addon, I start getting just a flat 0% for all files in the app, but I do get % for things in the in-repo addon.

It looks like this is due to the fact that multiple Istanbul plugins get added to the app config for apps with in-repo-addons, once for the app's `app` directory and once for the in-repo-addon's `app` directory.

For some reason I thought this wouldn't matter, but there must be a requirement to only have one Istanbul plugin per config. I suggest making a change to [_instrumentDirectory](https://github.com/kategengler/ember-cli-code-coverage/blob/v1.0.0-beta.1/index.js#L149-L159) so that it reuses an existing istanbul plugin if one exists: eg

```js
  _instrumentDirectory(appOrAddon, dir, modulePrefix) {
    if (existsSync(dir)) {
      let options = appOrAddon.options = appOrAddon.options || {};
      options.babel = options.babel || {};
      let plugins = options.babel.plugins = options.babel.plugins || [];
      let include = this._getIncludes(dir, modulePrefix);
      let plugin = plugins.find((plugin) => plugin[0] === this.IstanbulPlugin);
      if (plugin) { // plugin already exists, so amend it rather than adding another one
        plugin[1].include = plugin[1].include.concat(include);
      } else {
        plugins.push([this.IstanbulPlugin, {
          exclude: this._getExcludes(),
          include
        }]);
      }

    }
  },
```

I've tested this locally and confirmed that both app directories are being instrumented

```bash
----------------------------------|----------|----------|----------|----------|----------------|
File                              |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------------------|----------|----------|----------|----------|----------------|
All files                         |       75 |      100 |       40 |       75 |                |
 app                              |      100 |      100 |        0 |      100 |                |
  app.js                          |      100 |      100 |      100 |      100 |                |
  resolver.js                     |      100 |      100 |      100 |      100 |                |
  router.js                       |      100 |      100 |        0 |      100 |                |
 app/utils                        |       50 |      100 |       50 |       50 |                |
  my-covered-util-app.js          |      100 |      100 |      100 |      100 |                |
  my-uncovered-util-app.js        |        0 |      100 |        0 |        0 |              2 |
 lib/my-in-repo-addon/addon/utils |       50 |      100 |       50 |       50 |                |
  my-covered-util.js              |      100 |      100 |      100 |      100 |                |
  my-uncovered-util.js            |        0 |      100 |        0 |        0 |              2 |
 lib/my-in-repo-addon/app/utils   |      100 |      100 |      100 |      100 |                |
  my-covered-util.js              |      100 |      100 |      100 |      100 |                |
  my-uncovered-util.js            |      100 |      100 |      100 |      100 |                |
----------------------------------|----------|----------|----------|----------|----------------|
```

Hopefully, that will resolve #151 as well

rwwagner90:
@adamjmcgrath that does fit the issue of 0% coverage, however, now it is reporting 50% coverage, but I would expect it to be 83.33 or some random number, just like the other normal apps, since we do have app.js etc still.

rwwagner90:
Also, how did you get that nice table output of coverage like that?

adamjmcgrath:
> Also, how did you get that nice table output of coverage like that?

I added `config.reporters.push('text');` to https://github.com/kategengler/ember-cli-code-coverage/blob/master/lib/attach-middleware.js#L35

> I would expect it to be 83.33 or some random number, just like the other normal apps, since we do have app.js etc still.

That's weird, when I ran it I got total lines: 75%, which is roughly what I would expect. Could you run it with the text reporter and share the text report

rwwagner90:
@adamjmcgrath I tried adding the text reporter, but it doesn't seem to log to my console. Does the test need to pass?

rwwagner90:
Even with passing tests, the text reporter does not seem to log anything to the console.

rwwagner90:
@adamjmcgrath I got the table finally, here is my output:
```
----------------------------------|----------|----------|----------|----------|----------------|

File                              |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------------------|----------|----------|----------|----------|----------------|

All files                         |       50 |      100 |       50 |       50 |                |

 app                              |      100 |      100 |      100 |      100 |                |

  app.js                          |      100 |      100 |      100 |      100 |                |

  resolver.js                     |      100 |      100 |      100 |      100 |                |

  router.js                       |      100 |      100 |      100 |      100 |                |

 app/utils                        |      100 |      100 |      100 |      100 |                |

  my-covered-util-app.js          |      100 |      100 |      100 |      100 |                |

  my-uncovered-util-app.js        |      100 |      100 |      100 |      100 |                |

 lib/my-in-repo-addon/addon/utils |       50 |      100 |       50 |       50 |                |

  my-covered-util.js              |      100 |      100 |      100 |      100 |                |

  my-uncovered-util.js            |        0 |      100 |        0 |        0 |              2 |

 lib/my-in-repo-addon/app/utils   |      100 |      100 |      100 |      100 |                |

  my-covered-util.js              |      100 |      100 |      100 |      100 |                |

  my-uncovered-util.js            |      100 |      100 |      100 |      100 |                |

----------------------------------|----------|----------|----------|----------|----------------|
```

adamjmcgrath:
@rwwagner90 thanks, don't think my suggestion worked - will continue to investigate.

Noticed a couple of the imports in the in-repo-addon fixtures are incorrect fyi
https://github.com/kategengler/ember-cli-code-coverage/blob/9ad8a6c5b1e3e8b12011b788a9a06c6a5ad3c73d/test/fixtures/my-app-with-in-repo-addon/lib/my-in-repo-addon/app/utils/my-covered-util.js#L1
and
https://github.com/kategengler/ember-cli-code-coverage/blob/9ad8a6c5b1e3e8b12011b788a9a06c6a5ad3c73d/test/fixtures/my-app-with-in-repo-addon/lib/my-in-repo-addon/app/utils/my-covered-util.js#L1

(remove `addon` from the path)

adamjmcgrath:
I think there's an issue with the fact that

adamjmcgrath:
There seems to be an issue with:

https://github.com/kategengler/ember-cli-code-coverage/blob/9ad8a6c5b1e3e8b12011b788a9a06c6a5ad3c73d/test/fixtures/my-app-with-in-repo-addon/app/utils/my-uncovered-util-app.js#L1-L3

Having exactly the same content as:

https://github.com/kategengler/ember-cli-code-coverage/blob/9ad8a6c5b1e3e8b12011b788a9a06c6a5ad3c73d/test/fixtures/my-app-with-in-repo-addon/lib/my-in-repo-addon/addon/utils/my-uncovered-util.js#L1-L3

My, slightly wild, guess is that istanbul is hashing the content, then skipping files that try to be instrumented with the same content. I suggest you change `my-uncovered-util-app.js` to `myUncoveredUtilApp` (same with `my-covered-util-app.js`) 

I have to pop out now, but I'll take another look at this over the weekend

rwwagner90:
@adamjmcgrath 

>Noticed a couple of the imports in the in-repo-addon fixtures are incorrect fyi

Good call on the paths! I will fix them.

>My, slightly wild, guess is that istanbul is hashing the content, then skipping files that try to be instrumented with the same content. I suggest you change my-uncovered-util-app.js to myUncoveredUtilApp (same with my-covered-util-app.js)

There is definitely an issue if the files have the same name. Since ember will pull in the one from the addon's app directory and try to drop it into your app's namespace. I was seeing it only report for the in-repo addon, until I added `-app` to the names of the files in the app. I will play with some naming schemes and maybe completely change the files to see what works.

rwwagner90:
I tried changing the names, and that didn't help. It's very odd that this works on your machine and not mine.

adamjmcgrath:
It's not the file names, it's the file's content being the same that's the issue

adamjmcgrath:
I've done some more investigation and it looks like you can't use the Istanbul plugin multiple times with different configs. `babel-plugin-istanbul` effectively caches the include/exclude config (here: https://github.com/istanbuljs/babel-plugin-istanbul/blob/master/src/index.js#L17) this is why I was getting different coverage reports depending on which app folder I edited last.

To work around this I've updated the code to gather all the paths that should be included, then added them to every plugin. Istanbul just ignores the irrelevant paths so it seems to work fine.

I've put up a branch to demonstrate it here:
https://github.com/kategengler/ember-cli-code-coverage/compare/in-repo-addon-coverage...adamjmcgrath:in-repo-addon-coverage

When I run the integration tests I get: `AssertionError: expected 75 to equal 83.33` which I would expect and the generated html report looks good to me:

screen shot 2018-02-03 at 11 30 58

rwwagner90:
@adamjmcgrath are you on the Ember slack?

rwwagner90:
Tests are finally green, so going to merge this in. Thanks a ton @adamjmcgrath!

PR #159, opened 2018-02-04
fix babel-plugin-istanbul caching issue

rwwagner90:


rwwagner90:
@adamjmcgrath I am getting cannot read concat of undefined. https://travis-ci.org/kategengler/ember-cli-code-coverage/jobs/337094564#L691

I'm also a little confused on how this works. We are always passing an empty array to the plugins and only using the `includes` stuff internal to this addon ourselves? Are we using multiple plugins? I'm just not super familiar with Istanbul or this code, so sorry if these are obvious questions!

adamjmcgrath:
> I am getting cannot read concat of undefined. https://travis-ci.org/kategengler/ember-cli-code-coverage/jobs/337094564#L691

yep sorry my bad, I didn't fix the unit tests in that branch, I'll go ahead and do that today

> We are always passing an empty array to the plugins and only using the includes stuff internal to this addon ourselves?

We populate the `includes` array in `_instrumentAppDirectory`, `_instrumentAddonDirectory`, `_instrumentInRepoAddonDirectories`, then pass it to every plugin we use.

> Are we using multiple plugins? 

Yes, for an `addon` for example we add a plugin to the addon's options to instrument the addon directory. We also add a plugin to the addon's dummy app options to instrument the addon's app directory - we use the includes array so that we only instrument the addons app directory not the addon's dummy app directory. For an app with in-repo addons it get's more complicated, but it's the same premise (add the plugin to the app's options, and add the plugin to each of the in-repo-addon's options)

rwwagner90:
@adamjmcgrath ah okay. I'll trust this is all good once the tests pass then 👍 . Let me know when they are updated. Thanks so much for your help!

adamjmcgrath:
@rwwagner90 have fixed the tests and refactored the instrumentation logic a little, hopefully it's a bit clearer now.

rwwagner90:
Seems more or less okay, going to merge into my branch. Thanks @adamjmcgrath!

PR #163, opened 2018-02-14
Remove merge-coverage and explicit parallel option

adamjmcgrath:
The goal of this PR is to simplify measuring coverage when parallel testing, buy removing the `merge-coverage` command and explicit parallel options.

Instead, we can create a `coverageMap` once when we attach the middleware then merge the coverage data into it after each parallel test run. 

For the `ember serve` case, we would continue to create a new coverage map on every test run.


rwwagner90:
@adamjmcgrath so can you please explain this further? Does this essentially mean we can run `COVERAGE=true ember test` or `COVERAGE=true ember exam` and it just works magically with both cases?

adamjmcgrath:
Yep - exactly, this test should confirm that: https://github.com/kategengler/ember-cli-code-coverage/blob/a973cfe2f003aa8ae7e023ad6f3de61de4ba1aca/test/integration/app-coverage-test.js#L79-L88

rwwagner90:
@adamjmcgrath Nice! Will there be any performance issues with people running just plain `ember t`? We should probably document that parallel testing can now just be done with `COVERAGE=true ember exam` etc. That way, it is obvious it is supported.

adamjmcgrath:
> Will there be any performance issues with people running just plain ember t? 

No, the coverage merging logic is in the middleware which only get's attached when coverage is enabled

> We should probably document that parallel testing can now just be done with COVERAGE=true ember exam etc. That way, it is obvious it is supported.

👍 will update the README

rwwagner90:
@adamjmcgrath 
>No, the coverage merging logic is in the middleware which only get's attached when coverage is enabled

I mean when coverage is enabled. Will people running `COVERAGE=true ember t` experience any slowdowns or anything different than before? I'm basically wondering if we want to always run this stuff or if we should still pass a flag or check for `ember exam` before merging stuff.

adamjmcgrath:
> Will people running COVERAGE=true ember t experience any slowdowns or anything different than before?

`COVERAGE=true ember t` wont be affected at all - the coverage map will still be created once and the report will still be generated once.

kategengler:
I believe this will remove a bit of functionality: If you run ember exam partitioned (say in separate CI containers), you can no longer merge coverage into one report (https://github.com/trentmwillis/ember-exam#split-test-parallelization).

rwwagner90:
@kategengler I asked if this should work with CI still, and I thought @adamjmcgrath confirmed it would?

kategengler:
I believe it will work fine on CI if you're not using the partition option for ember exam.

rwwagner90:
@adamjmcgrath can you confirm if this will work with partition? I think you had said you were running with partition for your company's test suite right?

adamjmcgrath:
> I believe this will remove a bit of functionality: If you run ember exam partitioned (say in separate CI containers), you can no longer merge coverage into one report

Ah yep - I didn't consider this use case. I use the `parallel` feature of `ember exam` but not `partition`. We could put the `coverage-merge` command back although perhaps recommending something like `istanbul-merge` (or possibly `nyc` when istanbuljs/nyc/issues/515 is complete) would be simpler?

rwwagner90:
@adamjmcgrath I'll let you and @kategengler decide on that 😃. I don't use ember exam myself, so I thought parallel was the only thing we needed and that ran on multiple boxes, but I suppose I was confusing it with partition.

PR #164, opened 2018-02-21
Use the parent registry for determining JS extensions

dfreeman:
In combination with typed-ember/ember-cli-typescript#144, this should enable coverage reporting for TypeScript files. (Previously we were checking this addon's _own_ preprocessor registry, which would never contain any additional extensions)

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,10 +43,11 @@ module.exports = {
 
   // Ember Methods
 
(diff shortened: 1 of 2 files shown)


rwwagner90:
Sorry for the delay, great work @dfreeman!

PR #165, opened 2018-02-26
Add test-code-coverage command

mydea:
This PR adds a command `ember test-code-coverage`. This can be used e.g. in CI environments.

The command can also be used to simply print out the total coverage percentages, e.g. so you don't need to go looking into HTML files for them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -274,5 +274,11 @@ module.exports = {
     }
 
     return this._inRepoAddons;
+  },
+
+  includedCommands: function() {
+    return {
+      'test-code-coverage': require('./lib/commands/test-code-coverage')
+    };
   }
 };
--- a/lib/commands/test-code-coverage.js
+++ b/lib/commands/test-code-coverage.js
@@ -1,0 +1,160 @@
+'use strict';
+
+let fs = require('fs');
+let path = require('path');
+let chalk = require('chalk');
+let RSVP = require('rsvp');
+
+function camelize(str) {
+  str = str.split('-').map((str) => str.charAt(0).toUpperCase() + str.slice(1)).join('');
+  return str.charAt(0).toLowerCase() + str.slice(1);
+}
+
+module.exports = {
+
+  name: 'test-code-coverage',
+  description: 'Test the generated code coverage against given goals.',
+  works: 'insideProject',
+
+  availableOptions: [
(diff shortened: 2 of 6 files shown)


KillaPL:
Something like this would be very nice, gitlab can parse tests output and make a good use of coverage data out of the box, but it must be provided by the tests command.

rwwagner90:
@kategengler @adamjmcgrath what do you guys think about this feature?

index.js
```diff
@@ -274,5 +274,11 @@ module.exports = {
     }
 
     return this._inRepoAddons;
+  },
+
+  includedCommands: function() {
```
kategengler: I think this may need to be rebased; there's already an `includedCommands` hook

index.js
```diff
@@ -274,5 +274,11 @@ module.exports = {
...
+    return {
+      'test-code-coverage': require('./lib/commands/test-code-coverage')
```
kategengler: I prefer we name it `code-coverage:test` or `code-coverage-test`

rwwagner90:
@kategengler aside from the need for a rebase, are you in favor of this feature in general?

lib/commands/test-code-coverage.js
```diff
@@ -0,0 +1,160 @@
...
+   * Tries to extend `availableOptions` with globals from config.
+   *
+   * @public
+   * @method beforeRun
+   * @return {Void}
+   */
+  beforeRun() {
+    this._super.apply(this, arguments);
+
+    // try to read global options from `config/coverage.js`
```
kategengler: Use the util at lib/config.js to get the configuration and and the project's configPath for the location of the config (see https://github.com/kategengler/ember-cli-code-coverage/blob/master/index.js#L142)

rwwagner90:
@kategengler I also think we may want to defer this to nyc. @adamjmcgrath what do you think?

adamjmcgrath:
My preference would be nyc support and improving that tool, but I can't offer any help with that at the moment, so don't mind if you want to proceed with this solution.

rwwagner90:
I agree. I think we should perhaps close, in favor of nyc.

csprocket777:
ummm. what is `nyc`?

PR #166, opened 2018-02-27
Revert "Remove merge-coverage and explicit parallel option (#163)"

adamjmcgrath:
This reverts commit 0592f5f9cc861c18b889c1d11622dbb7678f61b3.

Reverting because PR #163 removed support for merging coverage reports generated after separate test runs, eg when using the `partition` feature of `ember exam`

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Configuration is optional. It should be put in a file at `config/coverage.js` (`
 
 - `coverageFolder`: Defaults to `coverage`. A folder relative to the root of your project to store coverage results.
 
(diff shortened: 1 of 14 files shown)


kategengler:
@adamjmcgrath Sorry for letting this linger. I think the improvements from the work this reverts to combine coverage as it comes in are great, but we just need to also have the parallel option and coverage-merge command for those using it in a partition. Do you want to update this PR to just bring back those parts?

adamjmcgrath:
> Do you want to update this PR to just bring back those parts?

Sure - no problem. Will do that this week

kategengler:
Thanks for bringing it back!

PR #190, opened 2018-09-07
Update to Ember 3.4, dropping support for Node 4

rwwagner90:


rwwagner90:
Would enhancement and breaking be the correct labels?

kategengler:
I think just one label, I'd go with breaking.

PR #212, opened 2019-02-25
ember-cli-typescript v2 fix

lifeart:
image

PR #223, opened 2019-10-18
Support Typescript / ember-cli-typescript

lifeart:


rwwagner90:
@lifeart this would close #212 right? Seems good to me

lifeart:
@rwwagner90 yes :)
// cc @makepanic, @NullVoxPopuli, @luketheobscure, @mdeanjones, @gmurphey, @scottkidder, @kategengler

NullVoxPopuli:
This does not work for me. Typescript files are ignored. :(

image

lifeart:
@NullVoxPopuli, plz retry) small typo fixed

NullVoxPopuli:
that did it!
image

index.js
```diff
@@ -196,12 +196,14 @@ module.exports = {
    * @returns {Array<String>} include paths
    */
   _getIncludesForDir: function(dir, prefix) {
+    const hasEmberCliTypescript = this.project && this.projectFindAddonByName && this.project.findAddonByName('ember-cli-typescript');
```
makepanic: this seems like a typo?

Shouldn't the second check be `this.project.findAddonByName`?

lifeart: @makepanic nice catch! Fixed

makepanic: Oops didn't see that you've already fixed it. Had the same issue nullvox initially reported and saw that in a diff between the other PR.

NullVoxPopuli:
I'm confused how some of these are untested : https://codeclimate.com/github/NullVoxPopuli/emberclear/code?sort=test_coverage

Resolver is used by ember itself.
Logout is for sure tested. :/
Am confused

lifeart:
image

makepanic:
Can confirm that with the fixed typo `ts` coverage works again.

lifeart:
@NullVoxPopuli it can be `pods` issue (don't related to `.ts` itselft) 
I can't click on uncovered file link (404)

NullVoxPopuli:
Yeah, I think I need to update how I upload the coverage since I also use a monorepo

config/ember-try.js
```diff
@@ -1,80 +1,76 @@
...
-const getChannelURL = require('ember-source-channel-url');
+const getChannelURL = require("ember-source-channel-url");
 
 module.exports = function() {
   return Promise.all([
-    getChannelURL('release'),
-    getChannelURL('beta'),
-    getChannelURL('canary')
-  ]).then((urls) => {
+    getChannelURL("release"),
```
rwwagner90: @lifeart were all these quotes changed to double quotes on purpose? I thought the blueprint used single quotes?

lifeart: @rwjblue it can be an `eslint` or `prettier` auto format rule

test/unit/index-test.js
```diff
@@ -496,11 +496,11 @@ describe('index.js', function() {
         it('instruments the inrepo addon', function() {
           const includes = Index._getIncludesForInRepoAddonDirectories();
           expect(includes).to.deep.equal([
-            'my-app/utils/my-covered-util.js',
```
rwwagner90: Do we no longer support folders for the app, in-repo-addon, etc? I think we should, right?

config/ember-try.js
rwwagner90: We should use whatever the default blueprint format is. I think it is single quote in all the Ember repos.

test/unit/index-test.js
lifeart: have no idea(
This changes just to get tests passed

rwwagner90: Hmm, maybe something changed in the resolver or something. I don't think we want to change the paths though.

lifeart: how we can fix tests in this case?

makepanic: I tried to investigate this change and it seems like the test is green if i run `yarn node-test` locally in `code-coverage-upstream-oct-2019` with changes in `test/unit/index-test.js` reverted :thinking: 

It even fails if i keep the changed index-test.js 



I tried this in both node8 and node latest

lifeart: I suppose it was failing on my windows machine, anyway changes reverted

package.json
```diff
@@ -22,65 +22,67 @@
...
     "loader.js": "^4.7.0",
     "mocha": "^5.2.0",
-    "qunit-dom": "^0.7.1",
+    "qunit-dom": "^0.8.4",
     "rimraf": "^2.6.2",
     "sinon": "^4.2.2"
   },
   "engines": {
-    "node": "6.* || 8.* || >= 10.*"
+    "node": "8.* || >= 10.*"
```
Turbo87: FYI this should be considered a breaking change

config/ember-try.js
```diff
@@ -7,31 +7,23 @@ module.exports = function() {
...
-          name: 'ember-lts-2.18',
-          env: {
-            EMBER_OPTIONAL_FEATURES: JSON.stringify({ 'jquery-integration': true }),
-          },
+          name: 'ember-lts-3.8',
           npm: {
             devDependencies: {
-              '@ember/jquery': '^0.5.1',
-              'ember-source': '~2.18.0'
+              'ember-source': '~3.8.0'
```
Turbo87: same for this, it looks like we're dropping support for old Ember versions

Issue #232, opened 2019-11-15
TypeError: Cannot convert undefined or null to object


ventuno:
Similarly to https://github.com/kategengler/ember-cli-code-coverage/issues/181 this error is visible when running `COVERAGE=true ember test --path=dist` or `COVERAGE=true ember exam --path=dist`.

This seems to happen because https://github.com/kategengler/ember-cli-code-coverage/pull/182/ fixes the problem by defaulting `fileLookup` in the `included` hook which is only called when `COVERAGE=true ember build ...` runs.

When using the `--path` argument, `build` does not run hence `fileLookup` is `null` again.

PR #182, opened 2018-06-01
Fix fileLookup is null in testemMiddleware

Gaurav0:
Fixes https://github.com/kategengler/ember-cli-code-coverage/issues/181

PR #233, opened 2019-11-15
Default fileLookup to {} instead of null.

ventuno:
Attempt to fix https://github.com/kategengler/ember-cli-code-coverage/issues/232.
See a detailed description and analysis of the issue in #232.

The idea is to default `fileLookup` to `{}` instead of `null` this way regardless of the fact that `included` will run or not `fileLookup` should always be non-null.

I'm not 100% sure if this is the best fix for this issue. The issue may be elsewhere in the addons system. I'll leave the reviewer(s) weigh in on this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ function getPlugins(appOrAddon) {
 // Regular expression to extract the file extension from a path.
 const EXT_RE = /\.[^\.]+$/;
 
-let fileLookup = null;
+let fileLookup = {};
 
 module.exports = {
   name: require('./package').name,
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -31,7 +31,7 @@ function getPlugins(appOrAddon) {
 // Regular expression to extract the file extension from a path.
 const EXT_RE = /\.[^\.]+$/;
 
-let fileLookup = null;
+let fileLookup = {};
```
rwjblue: Hmm, I actually think that #182 was not the right path forward here. Keeping this in module scope means we don't support running multiple projects within the same process.

Can you tweak things to instead revert the changes in #182 (which will bring back `this.fileLookup`) and then initialize it in `init`? It will have the same basic result as this PR does now (and fix the issue that #182 was also trying to fix), but will avoid that downside...

ventuno: Makes total sense, done. Can you please take another look?

PR #236, opened 2020-01-27
Ensure coverage collection accounts for addons with a custom `moduleName` implementation

### LINKED ISSUE #235: Incorrect Coverage Reports when moduleName differs from package name
A comprehensive gist in reproducing the steps exists here:
https://gist.github.com/axelerate/541197497877db60ea416effa9dce5d8

I confirmed that this problem only manifests itself as an issue for addons where the package name is not the same as the value defined in index.js's `moduleName`. This thus leads to the coverage omitting most if not all files that were tested.

This problem first arose once our team migrated to Ember 3.8.3, given that we replaced the inline literal name of our addon from "bar" to require('./package').name, which reads out as "@foo/bar".

axelerate:
Invoking covered module name instead of the covered addon's project payload whenever assessing covered addon's name.
When referencing a covered addon, we should ideally use the module to get the module name, as this is the source of truth when designating the addon name. This fixes the issue wherein if there exists any discrepancy between the module and project name, the coverage report ends up being inaccurate.

Fixes #235

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,6 @@ var config = require('./lib/config');
 const walkSync = require('walk-sync');
 const VersionChecker = require('ember-cli-version-checker');
 const concat = require('lodash.concat');
-const MODULE_NAME = 'ember-cli-code-coverage';
 
 function requireBabelPlugin(pluginName) {
   let plugin = require(pluginName);
@@ -36,10 +35,6 @@ let fileLookup = null;
 
 module.exports = {
   name: require('./package').name,
-
-  moduleName() {
-    return MODULE_NAME;
-  },
 
   /**
    * Look up the file path from an ember module path.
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -258,6 +266,18 @@ module.exports = {
     return this._coveredAddon;
   },
 
+  /**
+   * Find the module (if any) that's being covered.
+   * @returns {Module} the module under test
+   */
+  _findCoveredModule: function() {
```
rwjblue: I don't understand why we need this (we should use `_findCoveredAddon` directly)

index.js
```diff
@@ -258,6 +266,18 @@ module.exports = {
...
+    if (!this._coveredModule) {
+      this._coveredModule = require(this.project.findAddonByName(this.project.pkg.name).root);
```
rwjblue: We should definitely not be requiring an addon's entry point directly and making assumptions about it.

index.js
```diff
@@ -240,7 +246,9 @@ module.exports = {
    */
   _parentName: function() {
     if (this.parent.isEmberCLIAddon()) {
-      return this._findCoveredAddon().name;
+      const coveredModule = this._findCoveredModule();
+
+      return coveredModule.moduleName ? coveredModule.moduleName() : coveredModule.name;
```
rwjblue: Let's revert these changes (though `_parentName` is unused)

PR #237, opened 2020-01-28
Merge pull request #237 from kategengler/rwjblue-patch-1

rwjblue:
Prior to this, the only thing that was being tested in CI was `lint:js`.


kategengler:
😱 Thanks for catching this!

Issue #240, opened 2020-02-05
Request to publish latest tag version of ember-cli-code-coverage


axelerate:
The latest code changes resolves ongoing issues when it comes to generating correct reports for our Ember addons.

mehulkar:
What are the blockers for releasing 1.0 (not beta)?

axelerate:
@kategengler Any blockers for release v1.0.0?

narayanpai:
@rwjblue Any updates on when the latest version can be released?

kategengler:
If anybody has time, before the next release I'd like to verify that the README and options presented in there are correct and working for the latest beta. Any volunteers?

axelerate:
I'll volunteer. Will post results.

rwwagner90:
@mehulkar there is an issue with things we need to do before 1.0.0 https://github.com/kategengler/ember-cli-code-coverage/issues/152

Plus I think there are concerns around the ever changing Ember build process, and that there may be a total rewrite of this addon at some point. @kategengler would know more about this and her plans for the future of this addon.

rwwagner90:
1.0.0-beta.9 has been published. Hopefully we can get an actual 1.0.0 out soon! I believe 0.4.2 doesn't work with modern Ember, and 1.0.0 has been in beta a long time, so my vote would be to go ahead and publish.

axelerate:
Thanks @rwwagner90 ! I see you've modified the version name, but I'm not seeing the version tag for 1.0.0-beta.9 being published. Am I missing something?

kategengler:
@axelerate Thanks for taking that on. Looking forward to hearing your results. 

fwiw, @rwwagner90, I personally use 0.4.2 with modern ember and have had no difficulties. It's one of the things that concerns me -- some combination of circumstances makes it not work for only some people.

rwwagner90:
@kategengler what's your project like? I hit several things, like I think there was first an issue with babel 6 or 7 updates, some things with glimmer 2 and now glimmer.js etc. I don't 100% know, but I think there are a lot of things incompatible with 0.4.2 and I have been using the 1.0.0 betas for years across multiple different organizations with varying setups, and had no issues, so I think the beta has proved it's working.

rajasegar:
@kategengler v0.4.2 doesn't work with modern Ember. I am using ember 3.17

rwwagner90:
@rajasegar please use 1.0.0-beta.9

[...]

Issue #220, opened 2019-07-16
Coverage not working for helpers


katieburke1:
We have a folder with 10 helpers, 6 of them are displayed as not being covered even though they are. Does anyone have any ideas as to why this might be?

Turbo87:
unfortunately this is a bit hard to debug without a reproduction project 😞

PR #249, opened 2020-05-13
[BREAKING] Octane upgrade v3.12.0...v3.16.2

rajasegar:
Fixes errors with the latest  class syntax for modern Ember apps.

Related #220 #240


rajasegar-c:
Related #220 #240

package.json
```diff
@@ -31,7 +31,8 @@
     "babel-plugin-transform-async-to-generator": "^6.24.1",
     "body-parser": "^1.19.0",
     "co": "^4.6.0",
-    "ember-cli-babel": "^7.11.0",
+    "ember-cli-babel": "^7.17.2",
+    "ember-cli-htmlbars": "^4.2.2",
```
rwjblue, line 35: Can you move this back into devDependencies? I do not believe that it is needed as a dep.

package.json
```diff
@@ -42,43 +43,48 @@
...
     "loader.js": "^4.7.0",
     "mocha": "^5.2.0",
-    "qunit-dom": "^0.8.4",
+    "qunit-dom": "^1.0.0",
     "rimraf": "^2.6.2",
     "sinon": "^4.2.2"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": "10.* || >= 12"
```
rwjblue, line 84: Just pointing out, this needs to be a breaking change release

rwjblue:
> Fixes errors with the latest class syntax for modern Ember apps.

What specifically is fixing these errors? The ember-cli-babel update?

package.json
rajasegar: Done

package.json
rajasegar: Thanks Robert, will update the PR subject for the same.

rajasegar:
> > Fixes errors with the latest class syntax for modern Ember apps.
> 
> What specifically is fixing these errors? The ember-cli-babel update?

Yes, @rwjblue there were some errors related to class syntax I remember, but I am not quite sure about that If I use the 1.0 latest beta, it was working fine. Anyways I thought for compatibility reasons it's better to be on the latest LTS

rwjblue:
>  Anyways I thought for compatibility reasons it's better to be on the latest LTS

Totally agree!!!

PR #251, opened 2020-05-21
fix `InvalidStateError` exception

### LINKED ISSUE #244: The value is only accessible if the object's 'responseType' is '' or 'text' (was 'json')
## Error log

```
Uncaught DOMException: Failed to read the 'responseText' property from 'XMLHttpRequest': The value is only accessible if the object's 'responseType' is '' or 'text' (was 'json').
```

As suggested from the error log, `foo.responseText` was called, but `foo.responseType === json`.

## How to reproduce?

This should happen whenever `REQUEST_ASYNC === true && request.readyState === 4`

https://github.com/kategengler/ember-cli-code-coverage/blob/046e445b47e4a5da0aa5ea4d27380d9288cbedd0/lib/templates/test-body-footer.html#L22-L29

## Reason

https://github.com/kategengler/ember-cli-code-coverage/blob/046e445b47e4a5da0aa5ea4d27380d9288cbedd0/lib/templates/test-body-footer.html#L53-L56

When `responseType === 'json'`, the `handleCoverageResponse` is still called and `xhr.responseText` would cause an error, and halt the whole test.

```
ok 138 Chrome 80.0 - [0 ms] - ember-qunit: Ember.onerror validation: Ember.onerror is functioning properly
```

### Function that handles the `/write-coverage` and sends the data without response property
https://github.com/kategengler/ember-cli-code-coverage/blob/d6e2262923e7caeeb383fb6ccdf7b46e227f5715/lib/attach-middleware.js#L75-L78

### Function that creates the data sent as a response
https://github.com/kategengler/ember-cli-code-coverage/blob/d6e2262923e7caeeb383fb6ccdf7b46e227f5715/lib/attach-middleware.js#L57

brokenalarms:
fix `InvalidStateError` exception

This commit fixes #244.

The xhr.responseText property cannot be accessed unless xhr.responseType is either 'text' or '',
otherwise it throws an error:

Uncaught InvalidStateError: Failed to read the 'responseText' property from 'XMLHttpRequest':
The value is only accessible if the object's 'responseType' is '' or 'text' (was 'json')

This will therefore happen in handleCoverageResponse whenever
a request returns falsy. This should never happen, but is occasionally
flaking and producing the above error.

This commit fixes a few likely culprits:
- responseType = `json` was being set after `xhr.send`, which may cause a race condition
  https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/response
- the middleware called res.send with a CoverageMap may receive a non-serializable
  object, since istanbul CoverageMap provides a toJSON method, implying that
  the default method may not be serializable
- removing now-unused PhantomJS code.
- finally, introduces a check to only read from the 'responseText'
property in a valid circumstance, and gives a more meaningful error otherwise.

### DIFF
--- a/lib/attach-middleware.js
+++ b/lib/attach-middleware.js
@@ -54,7 +54,7 @@ function reportCoverage(map, root, configPath) {
 function coverageHandler(map, options, req, res) {
   writeCoverage(req.body, options.fileLookup, options.root, map);
   reportCoverage(map, options.root, options.configPath);
-  res.send(map.getCoverageSummary());
+  res.send(map.getCoverageSummary().toJSON());
 }
 
 // Used when app is in dev mode (`ember serve`).
--- a/lib/templates/test-body-footer.html
+++ b/lib/templates/test-body-footer.html
@@ -1,7 +1,4 @@
 <script>
-  // Make synchronous requests in phantom to avoid killing the runner before the coverage completes.
-  var REQUEST_ASYNC = !/PhantomJS/.test(window.navigator.userAgent);
-
   function sendCoverage(callback) {
     {%ENTRIES%}.forEach(function(file) {
       try {
@@ -17,21 +14,14 @@
     var request = new XMLHttpRequest();
     request.open('POST', '/write-coverage', REQUEST_ASYNC);
     request.setRequestHeader('Content-Type', 'application/json; charset=utf-8');
+    request.responseType = 'json';
     request.send(data);
 
-    if (REQUEST_ASYNC) {
-      request.responseType = 'json';
-      request.onreadystatechange = function() {
-        if (request.readyState === 4) {
-          handleCoverageResponse(this, callback);
-        }
-      };
-    } else {
-      // The request is already done at this point, since it is synchronous
-      if (request.status === 200) {
-        handleCoverageResponse(request, callback);
+    request.onreadystatechange = function() {
+      if (request.readyState === 4) {
+        handleCoverageResponse(this, callback);
       }
-    }
+    };
   }
 
   if (typeof Testem !== "undefined" && Testem.afterTests) {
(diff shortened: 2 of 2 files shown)


rwjblue:
I guess the thing I'm curious about is why is the response type `text` or empty?

brokenalarms:
It won't be. It's just to guard against the case where xhr rejects or returns falsy by expressing the only satisfactory conditions on which `xhr.responseText` property should ever be read.

When this happens, and the content type is `json` (as it always is), this will cause the `xhr.responseText` to be incorrectly read rather than just return the original falsy value of `xhr.response`.

brokenalarms:
Satisfying constraints of https://xhr.spec.whatwg.org/#the-responsetext-attribute
1 - this PR
2 - state is always `done` in this function
3 - already exists by falling back to text in the remaining valid scenarios.

rwjblue:
After poking at this a bit with @brokenalarms, I think we should do a few things:

1. Ensure that our middleware responds with the expected content type

https://github.com/kategengler/ember-cli-code-coverage/blob/d6e2262923e7caeeb383fb6ccdf7b46e227f5715/lib/attach-middleware.js#L54-L58

2. Remove phantomjs related codepaths and comments (💀)

https://github.com/kategengler/ember-cli-code-coverage/blob/81b57078d362b16dafa0197756e9ff8938151996/lib/templates/test-body-footer.html#L2-L3

https://github.com/kategengler/ember-cli-code-coverage/blob/81b57078d362b16dafa0197756e9ff8938151996/lib/templates/test-body-footer.html#L56-L59

3. Use a guard (similar to the one in this PR) to ensure we have the expected response type, and if we *don't* error

brokenalarms:
1. Looks like Express correctly sets to JSON if the provided object is JSON and we always assume we're dealing with `json` [http://expressjs.com/en/5x/api.html#res.send](here).

Theory is there's been a non-serializable object provided sometimes in `map.getCoverageSummary()` since `istanbul` provides additional [`toJSON` method](https://github.com/istanbuljs-archived-repos/istanbul-lib-coverage/blob/d8d12d40d1643d1f9ef4d96f80778d9a163f1e16/lib/coverage-map.js#L64) this infers that the `getCoverageSummaryMap()` alone may not always be serializable.

rwwagner90:
I defer to the wisdom of @rwjblue. @brokenalarms if you haven't already, can you adjust based on his feedback?

brokenalarms:
Have done, thanks!

rwwagner90:
Thanks for the PR @brokenalarms!

PR #262, opened 2020-05-27
Fix issue with coverage report.

rwjblue:
Unfortunately, this codepath is not under test and the changes from https://github.com/kategengler/ember-cli-code-coverage/pull/251 left it in an invalid state.

* The REQUEST_ASYNC variable was no longer present, and therefore referencing it threw an error
* The guard around `xhr.responseText` was no longer needed (we only ever get `JSON`, and `xhr.response` is already parsed properly for us)

We ultimately need to refactor to get this coverage reporting under test (plausibly an end-to-end acceptance test that confirms the expected JSON reporter output after running all the tests).

### DIFF
--- a/lib/templates/test-body-footer.html
+++ b/lib/templates/test-body-footer.html
@@ -12,7 +12,7 @@
     var data = JSON.stringify(coverageData || {});
 
     var request = new XMLHttpRequest();
-    request.open('POST', '/write-coverage', REQUEST_ASYNC);
+    request.open('POST', '/write-coverage');
     request.setRequestHeader('Content-Type', 'application/json; charset=utf-8');
     request.responseType = 'json';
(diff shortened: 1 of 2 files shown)


lib/templates/test-body-footer.html
```diff
@@ -12,7 +12,7 @@
     var data = JSON.stringify(coverageData || {});
 
     var request = new XMLHttpRequest();
-    request.open('POST', '/write-coverage', REQUEST_ASYNC);
+    request.open('POST', '/write-coverage');
```
rwjblue, line 15: XHR's `.open` method defaults to async if not given a value of `false` for the 3rd argument.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open

PR #263, opened 2020-05-27
Remove deprecated istanbul-api package.

rwjblue:
`istanbul-api` was deprecated in favor of using the individual `instanbul-*` libraries migrated to here.

---

Because we don't currently support running code coverage against a dummy app's code, and this addon doesn't have any code in addon/addon-test-support/app I can't actually verify if this is working very easily.

The things that I would like verified are:

* `COVERAGE=true ember test` emits a valid coverage file, and the reports look correct
* the coverage merge functionality still works properly (and can emit a valid final report)

### DIFF
--- a/lib/attach-middleware.js
+++ b/lib/attach-middleware.js
@@ -1,7 +1,9 @@
 'use strict';
 
 const bodyParser = require('body-parser').json({ limit: '50mb' });
-const istanbul = require('istanbul-api');
+const libCoverage = require('istanbul-lib-coverage');
+const libReport = require('istanbul-lib-report');
+const reports = require('istanbul-reports');
 const getConfig = require('./config');
 const path = require('path');
 const crypto = require('crypto');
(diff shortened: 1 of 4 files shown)


rwjblue:
Because we don't currently support running code coverage against a dummy app's code, and this addon doesn't have any code in addon/addon-test-support/app I can't actually verify if this is working very easily.

@kategengler / @rwwagner90 - Would one of y'all mind testing this against your applications to confirm things are working appropriately)/

rwwagner90:
@rwjblue happy to help test, but what exactly am I looking for? Just that coverage matches previous versions? I'm trying to identify where the gap in our tests here is and what would be different checking manually in an app. Just want to make sure I am looking for the right things!

kategengler:
@rwwagner90 See https://github.com/kategengler/ember-cli-code-coverage/pull/264 

We discovered the tests weren't running just after this PR.

rwwagner90:
@rwjblue I just tried this out and got `this._getIncludesForAddonDirectory is not a function or its return value is not iterable`

kategengler:
@rwwagner90 See https://github.com/kategengler/ember-cli-code-coverage/issues/265

rwwagner90:
@rwjblue @kategengler let me know when I should test this stuff out then. Happy to help when it's ready!

Turbo87:
I've rebased the commit and resolved the conflicts. CI is green now, so I'm assuming this is good to go?

rwwagner90:
@Turbo87 looks like there are some more conflicts. Once resolved, I would assume we could go ahead and merge.

rwjblue:
Rebased again.

PR #272, opened 2020-08-23
Fix `node-tests`

Turbo87:
Closes https://github.com/kategengler/ember-cli-code-coverage/pull/264

Unfortunately I had to completely remove the TypeScript tests since they were not working at all 🤷 

/cc @rwjblue


rwwagner90:
Hmm, I wonder why TS isn't working. @rwjblue any ideas? Thanks for fixing all these tests @Turbo87!

Turbo87:
there were several issues with the TS tests. after fixing some of them I have up. the latest issue was complaining about a missing babel dependency somewhere.

Turbo87:
merging this for now to unblock further work on this addon.

@lifeart @rwwagner90 since https://github.com/kategengler/ember-cli-code-coverage/pull/223 broke this and added TS support in the same PR it would be good if you could have a look at adding correctly working tests for this feature again to prevent it from breaking in the future.

rwwagner90:
@lifeart are you able to take a look? I don't use TS anywhere, so I am unfamiliar with what we need to do here.

lifeart:
TS is working fine

lifeart:
I don’t understand why tests is removed, even if they not running

lifeart:
I don’t understand why tests is removed, even if they not running

rwwagner90:
@lifeart what do you mean by it's working? I believe @Turbo87 is saying with these changes TS no longer works or at least the tests don't. We need some tweaks to get things working again on master.

lifeart:
@rwwagner90 tests not working, but coverage for ts working fine

rwwagner90:
@lifeart the tests not working implies to me TS doesn't work either, but again, I don't know about TS, so we need to poke at the master branch.

PR #278, opened 2020-08-30
mocha: Remove unnecessary Babel integration

Turbo87:
There does not appear to be any reason for us to use the Babel transpilation and polyfill, so this PR gets rid of them.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-    "presets": [ "@babel/preset-env" ]
-}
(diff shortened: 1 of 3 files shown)


rwwagner90:
@Turbo87 I do not know why we had this, but just thinking here, would this be needed if somehow you opted totally out of babel in your Ember app?

Turbo87:
no, this seems entirely unrelated to the Babel stuff in Ember itself

PR #289, opened 2020-09-17
Convert to Monorepo

Turbo87:
😱 😱 😱 😱 

no, wait, hear me out!

This PR is inspired by the repository layout that https://github.com/embroider-build/embroider/ is using. Instead of using the barely maintained https://github.com/tomdale/ember-cli-addon-tests project, we instead use a monorepo layout with several fixture packages in a `test-packages` folder. 

This enables us to use these test fixtures for debugging in a much easier way than before because they can also be used outside of running the test suite. In other words you can `cd` into the folder, run `COVERAGE=true ember test` there and manually verify the results in the `coverage` folder.

It also removes the need for running `npm install` during the test suite because the initial `yarn install` would have already installed all the necessary dependencies. Locally this seems to significantly speed up the test runs, for some reason they still seem to take roughly the same time on CI though.

PR #290, opened 2020-09-21
Use in-repo addon's "root" property for includes location

mdeanjones:
The goal is to instrument in-repo addons that don't reside directly within `/lib` by using the addon's root property, if available. 

Given how long the current implementation has been around I think I might be missing some important context that makes this less viable than a quick glance would suggest.


index.js
```diff
@@ -169,7 +169,11 @@ module.exports = {
    */
   _getIncludesForInRepoAddonDirectories() {
     return this._findInRepoAddons().reduce((acc, addon) => {
-      let addonDir = path.join(this.project.root, 'lib', addon.name);
+      let addonDir =
+        typeof addon.root === 'string'
+          ? addon.root
```
rwjblue, line 174: Hmm, why would `addon.root` not be present? I would think it always would...

index.js
```diff
@@ -169,7 +169,11 @@ module.exports = {
...
+          : path.join(this.project.root, 'lib', addon.name);
```
rwjblue, line 175: I agree, this seems super odd / not good. Assuming that an addon is in `lib/some-name` is definitely broken.

rwjblue (commented):
Overall, this definitely seems like the right direction! I'd like to land #289 first (to avoid that larger PR having too much rebase hell), but we should definitely get this in...

rwjblue:
OK, landed #289 so this is ready for a rebase. Note the testing system has changed a bit (not using `ember-cli-addon-tests` any more, for example), so the test might need to be tweaked a little more but overall I think the newer system is much easier to understand/work with...

mdeanjones:
Done, and might I add a very pleasant evolution to the testing story. The hesitance of touching anything `/lib` related was an overabundance of caution. It appears as though `addon.root` has been there since time immemorial but I didn't want to bork any edge-cases just in case.

rwjblue (approved):

rwjblue:
Thank you @mdeanjones!

test-packages/my-app/custom-dir-in-repo-addon.js
```diff
@@ -0,0 +1,78 @@
+'use strict';
+
+const path = require('path');
+const fs = require('fs-extra');
+const fixturify = require('fixturify');
+
+class InRepoAddon {
```
Turbo87, line 7: we don't need this file anymore, right?

apellerano-pw:
I have a question on how to exclude an in-repo-addon introduced into the coverage report by this PR.

We have a folder structure like
```
- client
  - our-app
  - design-system
```

And in our-app's package.json we specify the in-repo-addon with `../design-system` in the `ember-addon.paths` array.

How can I exclude this in-repo-addon from the coverage report? I've tried even `**/design-system/**` which i thought would be the glob catch-all but that didn't seem to do it.

PR #291, opened 2020-09-21
Use Jest for testing

Turbo87:
... instead of Mocha and Chai. And this introduces snapshot tests to assert on the `coverage-summary.json` results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,13 +9,13 @@
     "test-packages/*"
   ],
   "scripts": {
-    "test": "mocha test-packages/*-test.js"
+    "test": "jest --runInBand"
   },
   "devDependencies": {
     "chai": "^4.2.0",
     "chai-files": "^1.4.0",
     "execa": "^4.0.3",
-    "mocha": "^5.2.0",
+    "jest": "^26.4.2",
     "release-it": "^14.0.2",
     "release-it-lerna-changelog": "^2.4.0",
     "release-it-yarn-workspaces": "^1.5.0",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -9,13 +9,12 @@
     "test-packages/*"
   ],
   "scripts": {
-    "test": "mocha test-packages/*-test.js"
+    "test": "jest --runInBand"
```
rwjblue, line 12: Do we need in bad due to mutation of the local repo contents?

rwjblue (commented):
I think will need a rebase to account for the tests in https://github.com/kategengler/ember-cli-code-coverage/pull/290.

package.json
Turbo87: yep, and to prevent concurrent access of the git cleanup commands. there are probably ways around that, but for now this should be sufficient.

Turbo87:
rebase done ✅

rwjblue:
I think the `test-packages/index-test.js` also needs to be updated to Jest syntax:
```
FAIL test-packages/index-test.js
  ● index.js › _getIncludes › _getIncludesForInRepoAddonDirectories › for an app with an inrepo addon under a custom path › instruments the inrepo addon

    TypeError: Cannot read property 'deep' of undefined

      510 |         it('instruments the inrepo addon', function () {
      511 |           const includes = Index._getIncludesForInRepoAddonDirectories();
    > 512 |           expect(includes).to.deep.equal([
          |           ^
      513 |             'my-app/utils/my-covered-util.js',
      514 |             'my-app/utils/my-uncovered-util.js',
      515 |             'my-in-repo-addon/utils/my-covered-util.js',
```

Turbo87:
oh, whoops, totally missed that part 🤦‍♂️

rwjblue (approved):

PR #293, opened 2020-09-29
Fix empty reports when path option to existing build is used

robinborst95:
At the company where I work I was working on speeding up tests by splitting them and running them in parallel. For this I used the `--path` option, after which the coverage wasn't working anymore. I saw [this](https://github.com/kategengler/ember-cli-code-coverage/issues/271) issue, which is the exact problem I ran into. However, the initial fix mentioned there wasn't actually working, so the issue is still there in v1.0.0. Reason is that `!this.fileLookup` returns false for empty objects. After fixing that line, the coverage was working again.

I also added a basic test to make sure it actually works.

### DIFF
--- a/packages/ember-cli-code-coverage/index.js
+++ b/packages/ember-cli-code-coverage/index.js
@@ -99,7 +99,7 @@ module.exports = {
       return;
     }
 
-    if (!this.fileLookup) {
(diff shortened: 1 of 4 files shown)


GCheung55:
I tested this change on my project and there is coverage output when a path is given.

rwwagner90 (approved):

PR #294, opened 2020-10-07
Fix missing/incorrect file paths in lcov

GCheung55:
Given a repo with an in-repo-addon that re-exports a module, e.g. service, when the repo has an overriding file of what the in-repo-addon exports, then the overriding file should be included in the lcov file.

E.g., Given the following paths exists in the app
```
app/services/my-service.js
lib/my-in-repo-addon/addon/services/my-service.js
lib/my-in-repo-addon/app/services/my-service.js
```

Expected: lcov file should include the following lines.
```
SF:app/services/my-service.js
SF:lib/my-in-repo-addon/addon/services/my-service.js
```

Actual: lcov file includes the wrong path, such as the following lines.
```
SF:lib/my-in-repo-addon/app/services/my-service.js
SF:lib/my-in-repo-addon/addon/services/my-service.js
```

### DIFF
--- a/packages/ember-cli-code-coverage/index.js
+++ b/packages/ember-cli-code-coverage/index.js
@@ -128,8 +128,8 @@ module.exports = {
    */
   _getIncludes() {
     return [
-      ...this._getIncludesForInRepoAddonDirectories(),
       ...this._getIncludesForAddonDirectory(),
+      ...this._getIncludesForInRepoAddonDirectories(),
       ...this._getIncludesForAppDirectory(),
     ].filter(Boolean);
   },
--- a/packages/ember-cli-code-coverage/index.js
+++ b/packages/ember-cli-code-coverage/index.js
@@ -128,9 +128,9 @@ module.exports = {
    */
   _getIncludes() {
     return [
-      ...this._getIncludesForAppDirectory(),
-      ...this._getIncludesForAddonDirectory(),
       ...this._getIncludesForInRepoAddonDirectories(),
+      ...this._getIncludesForAddonDirectory(),
+      ...this._getIncludesForAppDirectory(),
     ].filter(Boolean);
   },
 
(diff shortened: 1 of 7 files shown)


GCheung55:
I thought of two solutions.
1. Only add the to `fileLookup` object if the key didn't already exist.
2. Reverse the order of how the paths are included, which gives `app/` paths a higher priority than the addon's `app/` paths.

GCheung55:
`test-packages/in-repo-addon-test.js` is failing because the snapshot is different. I'm not sure how to update the snapshot and if I should update the snapshot.

packages/ember-cli-code-coverage/index.js
```diff
@@ -128,9 +128,9 @@ module.exports = {
    */
   _getIncludes() {
     return [
-      ...this._getIncludesForAppDirectory(),
-      ...this._getIncludesForAddonDirectory(),
       ...this._getIncludesForInRepoAddonDirectories(),
+      ...this._getIncludesForAddonDirectory(),
```
rwjblue, line 132: The ordering between in-repo addons and addons is a bit tricky here. I don't think we can guarantee that in-repo addons are "clobbered" by "normal" addons (in fact I think I'd expect the opposite)... 🤔 

What do you think about using:

```suggestion
      ...this._getIncludesForAddonDirectory(),
      ...this._getIncludesForInRepoAddonDirectories(),
```

rwjblue (commented):
Thanks for working on this! Only had one small comment, but overall this looks good to me!

packages/ember-cli-code-coverage/index.js
GCheung55: That makes sense to me. I'd expect that the in-repo-addon would have a higher priority than an addon. I'll make the switch.

rwjblue:
> › 1 snapshot failed from 1 test suite. Inspect your code changes or run `yarn test -u` to update them.

@GCheung55 - Can you do `yarn test -u` to update the snapshot with the new coverage data?

rwjblue (approved):
Thanks!

Issue #300, opened 2020-12-16
README not easily discoverable in v1.0.1 and v1.0.2


ijlee2:
## Description

Hello. I wanted to learn how to install and configure `ember-cli-code-coverage` today. I noticed that the `master` branch did not contain the README file in the root folder, and ended up switching to the `v1.0.0` tag to find the README.

It seemed like the project structure might have changed in `v1.0.1`. I was wondering if it can be still possible to have a README in the root folder so that visitors can easily find it. The README in the root folder could be a verbatim copy or display a link to the README in `packages/ember-cli-code-coverage`.

Another option may be to add the link to README under the About section?

<img width="400" alt="About section in GitHub repo site" src="https://user-images.githubusercontent.com/16869656/102362416-f8a3a900-3f79-11eb-83a0-30e575025a14.png">

kategengler:
I think the README should be moved top level and it was just missed in the reorganization. 

@rwjblue Can you think of any reason not to do that?

chrisvdp:
@ijlee2 @kategengler we can close this now.

ijlee2:
@chrisvdp Thank you for your help!

PR #301, opened 2021-01-18
move readme to root

chrisvdp:
It seems the repo readme got moved into a folder when the repo was reorganized and it means users landing in the repo are presented with no docs

#300


kategengler:
Thank you!

rwjblue:
Hmm, I think we actually need to tweak this a bit. As it stands, the npm page looks a bit odd:

https://www.npmjs.com/package/ember-cli-code-coverage

PR #306, opened 2021-03-08
Allow to send coverage in development

krasnoukhov:
These changes aim to add support for coverage measurement when running in dev mode. The use case I had for this is making integration specs (which are using Capybara) collect and report frontend app coverage.

The summary of the PR is:
- Add an alias for `content-for` (called `code-coverage-footer`) so it can be used in non-test index.html without additional content that comes with ember-cli
- Change `sendCoverage` to handle JSON response better and make it not emit integration warning for non-test Ember env


krasnoukhov:
No longer relevant.

PR #327, opened 2021-10-12
DO NOT MERGE: Adds Embroider support + restructure

thoov:
This PR is primarily for review and will most likely be broken into smaller PRs once we try to land any of these changes.

Changes:

Addon is now consumed like:

ember-cli-build.js: 
```js
let app = new EmberApp(defaults, {
  babel: {
    plugins: [
      require('ember-cli-code-coverage').buildBabelPlugin({
        project: defaults.project,
      }),
    ],
  },
});
```

tests/index.html:
```html
<script src="/coverage.js" integrity="" data-embroider-ignore></script>
```


packages/ember-cli-code-coverage/index.js
```diff
@@ -1,85 +1,23 @@
...
-        )
-      );
+    if (opts.embroider === true) {
+      let { stableWorkspaceDir } = require('@embroider/compat/src/default-pipeline');
+      cwd = stableWorkspaceDir(project.root);
     }
 
-    return undefined;
+    const IstanbulPlugin = require.resolve('babel-plugin-istanbul');
+    return [IstanbulPlugin, { cwd, include, exclude }];
```
rwjblue, line 20: Let's make the consumer do `plugins: [...buildBabelPlugin()]` just in case we need to begin emitting multiple plugins in the future...

packages/ember-cli-code-coverage/index.js
```diff
@@ -1,85 +1,23 @@
...
-    this._super.included.apply(this, arguments);
-
-    this.parentRegistry = appOrAddon.registry;
-
-    if (!this._registeredWithBabel && this._isCoverageEnabled()) {
-      let checker = new VersionChecker(this.parent).for(
-        'ember-cli-babel',
-        'npm'
-      );
+  buildBabelPlugin(opts = {}) {
```
rwjblue, line 8: Let's add documentation to this (with a simple `@example` for how we expect folks to use it in `ember-cli-build.js` and for addons) kinda thing

packages/ember-cli-code-coverage/index.js
```diff
@@ -1,85 +1,23 @@
...
+    let project = opts.project;
+    let cwd = opts.cwd || project.root;
+    let include = opts.include || '**/*';
+    let exclude = opts.exclude || ['*/mirage/**/*'];
```
rwjblue, line 12: Let's add a short circuit return here if `process.env.COVERAGE` is falsey, this brings.
```suggestion
  buildBabelPlugin(opts = {}) {
    if (process.env.COVERAGE) {
      return [];
    }

    let project = opts.project;
    let cwd = opts.cwd || project.root;
    let include = opts.include || '**/*';
    let exclude = opts.exclude || ['*/mirage/**/*'];
```

packages/ember-cli-code-coverage/index.js
```diff
@@ -1,85 +1,23 @@
...
-    this.fileLookup = {};
-  },
-
-  included(appOrAddon) {
-    this._super.included.apply(this, arguments);
-
-    this.parentRegistry = appOrAddon.registry;
-
-    if (!this._registeredWithBabel && this._isCoverageEnabled()) {
-      let checker = new VersionChecker(this.parent).for(
```
rwjblue, line 40: I think we can probably delete this dependency now?

packages/ember-cli-code-coverage/addon-test-support/index.js
```diff
@@ -0,0 +1,27 @@
+let __require;
```
rwjblue, line 1: Can you add a bit of documentation to the top of this file explaining the conceptual thing that it is doing?

packages/ember-cli-code-coverage/addon-test-support/index.js
```diff
@@ -0,0 +1,27 @@
...
+  __require = __webpack_require__;
+}
+
+export function forceModulesToBeLoaded() {
+  // check if webpack is present
+  if (typeof __webpack_require__ !== 'undefined') {
+    let __modules = __require.m;
+    Object.keys(__modules).forEach(moduleName => {
+      if (!moduleName.startsWith('../') && !moduleName.startsWith('./node_modules/')) {
+        __require(moduleName);
```
rwjblue, line 13: Let's put this in a try / catch; we don't want errors to cause failures

packages/ember-cli-code-coverage/addon-test-support/index.js
```diff
@@ -0,0 +1,27 @@
...
+      }
+    });
+  }
+
+  Object.keys(window.requirejs.entries).forEach(moduleName => {
+    try {
+      if (!moduleName.includes('ember-test-app/tests')) { // this causes an infinite loop without it
```
rwjblue, line 20: Why does this infinite loop?

villander:
@thoov @rwjblue and how about the standalone engines here?

PR #343, opened 2021-11-22
Adding `modifyAssetLocation` config option

thoov:
Allow customizing how the file path adjustment works by passing a function as an option. Removing the namespace mapping override as its not really needed with the customAdjustment function.


thoov:
cc: @rwjblue

packages/ember-cli-code-coverage/lib/attach-middleware.js
```diff
@@ -81,6 +86,14 @@ function adjustCoverageKey(root, filepath, namespaceMappings) {
     pathWithoutNamespace = pathWithoutNamespace.slice(1);
   }
 
+  if (customAdjustment) {
+    let customPath = customAdjustment(root, relativePath, filepath);
```
rwjblue, line 90: Should we pass the namespaceMapping map also?

rwjblue (approved):

rwjblue (approved):

PR #345, opened 2021-12-01
Add test for files in app directory of in-repo addons

robinborst95:
This PR adds a test to confirm `modifyAssetLocation` works as expected when not using the `--path` option. I cherry-picked this from https://github.com/kategengler/ember-cli-code-coverage/pull/342 and used `modifyAssetLocation` to make the tests pass. The example in the readme wasn't sufficient though, as the `relativePath` doesn't start with `app`, but with the project name (as this is not replaced by the `namespaceMapping` yet at this point), so I adjusted the readme to make this more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,10 +211,12 @@ be relative to your project root.
 const app = new EmberApp(defaults, {
   'ember-cli-code-coverage': {
     modifyAssetLocation(root, relativePath) {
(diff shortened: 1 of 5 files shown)


rwjblue:
Thank you!!!

rwjblue (approved):

PR #366, opened 2022-11-17
Added ability to pass coverage api value and report type from client app config

saloniy:
Added the ability to pass the value of the API from the client application instead of using `write-coverage` always [as discussed](https://github.com/kategengler/ember-cli-code-coverage/issues/147#issuecomment-1302348256).
The default will be set to `write-coverage` just in case the value is not passed from the client app.
Tested this with my app and it's working as expected after adding the `coverageApiPath` in my app's config/coverage.js
image


krasnoukhov:
This looks useful to me

jagthedrummer:
I've run into the problem where the addon doesn't work with `ember-electron` and I'd love to see this fix get merge.

PR #378, opened 2023-03-09
remove unused @embroider/core dependency

mansona:
I don't really see where `@embroider/core` is being used even though it's a dependency of this package 🤔 I have run the tests after removing the dependency and it seems to work fine 🎉 

I noticed this when I was trying to try out the new unstable release of embroider and pnpm wasn't allowing me to update because it didn't match the version defined in this package. I figured if it's not being used the best solution is to remove it 👍


mansona:
Turns out it's not as simple as that 😂 `@embroider/core` is a peer dependency so I have to try this a different way

PR #379, opened 2023-03-09
allow for a wider range of embroider dependencies

mansona:
This is my second attempt at fixing my tests downstream. #378 wasn't enough of a change

This PR should be ok to merge and release as a minor and shouldn't interfere with any embroider dependencies dowstream.

**Note:** I needed to re-role the yarn.lock file in this PR because there are some babel changes that were causing some issues and the primary advice seems to be to re-roll your lock file 🤷 https://github.com/babel/babel/discussions/12949

I also had to update snapshots since there was some slight change. I'm assuming this is just as a result of the yarn.lock bump 👍

### DIFF
--- a/packages/ember-cli-code-coverage/package.json
+++ b/packages/ember-cli-code-coverage/package.json
@@ -34,8 +34,8 @@
     "test:ember-compatibility": "ember try:each"
   },
   "dependencies": {
-    "@embroider/compat": "^0.47.0",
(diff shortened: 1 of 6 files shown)


rwjblue:
In general, this seems good to me. I'm not sure why CI didn't run though 🤔

mansona:
This is fixed in https://github.com/kategengler/ember-cli-code-coverage/pull/382 👍

Issue #386, opened 2023-08-08
Coverage not collected in Embroider >3.1


chrismllr:
Versions:
```
"ember-cli": "~4.12.1",
"@embroider/compat": "^3.2.1",
"@embroider/core": "^3.2.1",
"@embroider/webpack": "^3.1.5",
"ember-cli-code-coverage": "2.0.0"
```

Utilizing embroider >2 (as of writing, 3.2) coverage is not generated. 

From initial debugging, a couple things ive noticed:

- The `stableWorkspaceDir` (@embroider/compat) method returns a directory that does not match the build directory that is printed when the build completes before tests are run (`Built in ....`) -- **Unsure if this is actually an issue, or Istanbul simply needs an empty directory to work from**
- `testMiddleware` method is registered, but the app never POSTs back (`coverageHandler` is never invoked)
- If running with `--server` flag, `sendCoverage` is invoked, but `window.__coverage__` is empty

> Repro: https://github.com/chrismllr/repro-code-coverage-embroider

chrismllr:
Attempted to point to this version of the repo with no luck:
https://github.com/kategengler/ember-cli-code-coverage/pull/382#issuecomment-1558314297

chrismllr:
Still digging, but It looks like the this package uses `stableWorkspaceDir` to pass istanbul the correct directory (the one embroider will build into):

```js
if (opts.embroider === true) {
  let {
    stableWorkspaceDir,
  } = require('@embroider/compat/src/default-pipeline');
  cwd = stableWorkspaceDir(cwd, process.env.EMBER_ENV);
}
```

But in later versions of embroider, i'm unsure what is used to determine that build dir:
```js
// embroider/compat v1.8.3

function defaultPipeline(emberApp, packager, options = {}) {
    let outputPath;
    let addons;
    if (process.env.REUSE_WORKSPACE) {
        addons = new _1.PrebuiltAddons(emberApp, options, process.env.REUSE_WORKSPACE);
    }
    else {
        if (process.env.SAVE_WORKSPACE) {
            options.workspaceDir = process.env.SAVE_WORKSPACE;
        }
        else {
            options.workspaceDir = stableWorkspaceDir(emberApp.project.root); // <----------- Same method used here
        }
        emberApp.project.ui.write(`Building into ${options.workspaceDir}\n`);
        addons = new _1.Addons(emberApp, options);
        addons.ready().then(result => {
            outputPath = result.outputPath;
[...]

PR #387, opened 2023-08-09
support new embroider working dir introduced in embroider/compat >3.1

chrismllr:
Relates to #386 

Allow Istanbul to observe the correct files when embroider builds into its new working directory location.

As of this commit: https://github.com/embroider-build/embroider/commit/9e078f07d94eb1291875912637da4e802ec1339e, published in version 3.1.5 of `@embroider/compat`, Embroider uses a new `locateEmbroiderWorkingDir` method to obtain where the "rewritten" app is built to, which now resolves to `node_modules/.embroider/rewritten-app`.


NullVoxPopuli:
Do you have a sample app that proves that this works?

chrismllr:
@NullVoxPopuli Sure thing, I can see about duping the repro repo and applying these changes.

chrismllr:
Test run from reproduction repo (no coverage output): https://github.com/chrismllr/repro-code-coverage-embroider/actions/runs/5881955365/job/15951451547

Test run from fix repo (coverage output successfully): https://github.com/chrismllr/fix-code-coverage-embroider/actions/runs/5881976092/job/15951525068

packages/ember-cli-code-coverage/index.js
```diff
@@ -55,10 +55,23 @@ module.exports = {
     }
 
     if (opts.embroider === true) {
-      let {
-        stableWorkspaceDir,
-      } = require('@embroider/compat/src/default-pipeline');
-      cwd = stableWorkspaceDir(cwd, process.env.EMBER_ENV);
+      let version = pkgJSON['devDependencies']['@embroider/compat'];
+      let semver = require('semver');
```
chrismllr, line 59: This admittedly feels flimsy, but `this.project` was not available in scope here (to pass on to something like ember-cli-version-checker)

chrismllr: Perhaps a try/catch for importing/requiring the new `locateEmbroiderWorkingDir` method

kategengler:
cc @mansona for embroider support here (unless there is someone else I should ping?)

nickschot:
I think it may be necessary to bump the peerDep requirements on `@embroider/core` (?). Just tried this PR as a patch and ran into a case where it ended up using an older `@embroider/shared-internals` package on CI which didn't have the `locateEmbroiderWorkingDir` yet, breaking the build with a missing function error. Adding a resolution fixes it of course.

rogeraraujo90:
@chrismllr can you confirm if the current version is only working with embroider 1? I've tested with embroider 2 but no coverage was reported either. This is currently a blocker for me.

rogeraraujo90:
cc: @ef4

RobbieTheWagner:
@chrismllr I would love to get this in, but want to make sure it doesn't make things blow up. We probably at least need to guard `locateEmbroiderWorkingDir` right?

SergeAstapov:
@chrismllr @RobbieTheWagner would it worth adding a test Embroider app to ensure this addon works?

RobbieTheWagner:
@SergeAstapov there is one https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/tree/master/test-packages/my-embroider-app

I think we would need a couple with different embroider versions. Tests are currently failing for this embroider app, so if anyone has time to help with this, I would appreciate it!

chrismllr:
Hey all, apologies for the lack of updates here, I have been taking paternity leave 👶 

I can attempt to get in to get some tests added as time permits, but if there's a way for someone to pick up and run with it, I can assist in any way I can!

As for some questions:
> @chrismllr I would love to get this in, but want to make sure it doesn't make things blow up. We probably at least need to guard `locateEmbroiderWorkingDir` right?

Could you clarify? Currently, its behind a `semver.satisfies` , so hopefully it would only be accessed if the version of `embroider/core` was sufficient (that method was introduced within 3.1)


> @chrismllr can you confirm if the current version is only working with embroider 1? I've tested with embroider 2 but no coverage was reported either. This is currently a blocker for me.

Nothing beyond my local testing; I've got a test repo mentioned in the original issue, but this may require some tests as others have mentioned.

RobbieTheWagner:
I merged #398 so not sure if we still need this PR?

chrismllr:
> I merged #398 so not sure if we still need this PR?

Ah, does that PR use some other method to tell istanbul where to look for app files?

AmauryD:
@chrismllr  I've only added a test for embroider/compat > 3.1. But i haven't changed anything related to this issue.
https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/blob/b9b5f79a2686522e4a65494beb32c13e1a5fe39c/test-packages/my-embroider-app-test.mjs#L16

chrismllr:
> @chrismllr I've only added a test for embroider/compat > 3.1. But i haven't changed anything related to this issue.
> 
> https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/blob/b9b5f79a2686522e4a65494beb32c13e1a5fe39c/test-packages/my-embroider-app-test.mjs#L16

I've added a commit to this fork to enable the above test to run on 3.1.0, and looks like its also pulled in some pnpm-lock updates per changes made in the last merged PR. Is there anything necessary to do to test the actions to run?

chrismllr:
Currently working through the tests for 3.1, but all embroider tests for 0,47, 1.0, and 2.0 are working 🎉 

There look to be some discrepancies with how `fixturify-project` modifies the package.json vs how the code within the block I've edited in this PR reads `pkgJSON`. Not getting the same values.

packages/ember-cli-code-coverage/index.js
chrismllr: Going to migrate to use the above, saw some issues in tests with `fixtureify-project` usage (these values read were not the fixture libs output). This method also feels more stable. If the method exists, use it. Otherwise, use the original.

chrismllr:
Altered the logic to use presence of the method instead of semver/ introspection of package.json. All tests passing for me locally!

RobbieTheWagner (approved):

PR #390, opened 2023-10-05
Handle relative paths when fixing keys

wagenet:
We have a mono repo with v2-style in-repo addons and tests. For instance, we have `ROOT/addons/internal` and `ROOT/tests/internal`. The path relative to our tests end up looking like `../../addons/internal`. Without the new conditional, we'd end up hitting the fallback case which tries to make everything relative to `tests/internal/app` this would end up getting translated into `ROOT/tests/addons/internal` which is _not_ the correct path. I believe it should always be correct to pass relative paths through as-is, which is what I've done here. This solves the problem for our uses.

### DIFF
--- a/packages/ember-cli-code-coverage/lib/attach-middleware.js
+++ b/packages/ember-cli-code-coverage/lib/attach-middleware.js
@@ -70,17 +70,15 @@ function adjustCoverageKey(
 ) {
   let relativePath = path.relative(root, filepath);
 
-  // This lives in a directory outside of the current one, likely a monorepo.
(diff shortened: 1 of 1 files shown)


kiwiupover:
@wagenet what is up with the test failures?

wagenet:
@kiwiupover looks like there's a bug for the embroider case. I believe I've fixed it now, but we should note that the embroider case is actually currently testing that we get no coverage metrics! This certainly seems wrong, but should be addressed in another PR.

RobbieTheWagner (approved):

PR #398, opened 2023-11-09
Switch to pnpm for Workspace Management and Tests Refactoring.

AmauryD:
@RobbieTheWagner I started to work on the pnpm conversion. But i have a few questions considering the tests.

Should we consider keeping the test applications on Ember 3.x, or is it advisable to update them to at least the lowest LTS release (4.4)? 
Should we drop the support of old Embroider versions (<= 1.0.0) ? 
Perhaps we should explore alternative methods for testing these test-apps, allowing for isolation from each other, and potentially enabling concurrent launches with varying dependency versions. Similar to what ember-try does.

I tried adding the test packages in the workspace, but the tests for the Embroider app are failing 50% of the time. I'll try again later.


AmauryD:
Not quite what I had in mind. Despite my efforts to maximize the speed of the embroider tests. Replacing dependencies necessitates reinstalling them, which is still quite time-consuming. At least it's consistent.

AmauryD:
Here's the summary of my changes: 

- All 'apps' tests are done in a separate folder with a clean install. Which prevents leftovers from previous tests.
- Jest was replaced by Vitest which have the same APIs. All tests are run in parallel by default now. The change of test engine is personal and could be reverted easily.
- Test apps were upgraded to the latest version of ember 3.x
- Replaced old Yarn v1 by pnpm v8.

RobbieTheWagner (approved):
This is awesome, thank you!

PR #319, opened 2021-06-12
Template coverage

mukilane:
In continuation of #103

Issue #392, opened 2023-10-09
Issues with first-class component templates?


spuxx1701:
Hello there! It appears I'm having issues with collecting coverage from first-class component templates (`<template>`).

When running `ember test` with `COVERAGE=true`, Im getting the following error on the first `.gts` file:

```bash
.../components/misc/loading-indicator/loading-indicator.component.ts: Scope functions can only consist of a single return statement which returns an object expression containing references to in-scope valuesError: .../components/misc/loading-indicator/loading-indicator.component.ts: Scope functions can only consist of a single return statement which returns an object expression containing references to in-scope values
```

Is there a known incompatibility between `ember-cli-code-coverage` and `gjs/gts` files? Is there a workaround I'm not aware of?

veelci:
@spuxx1701 I have 2 non-typescript projects on ember 4.12 and I'm running into this issue as well in one of the projects.
I've aligned the dependencies and configuration of both projects to be the same, so I'm not entirely sure what the issue is.  The only thing that is different between the working and non-working project is that I migrated _all_ of the files to .gjs at once in the working project.

vstefanovic97:
Hi, I've been trying to get `.gjs/.gts` coverage working for a few days, I've done some digging and I think I've Identified the steps to get this working.

So there are a couple of things needed.
1. Add `.gjs/.gts` extension to `babel-plugin-istanbul `'s extension options (this is just so that it recognizes and instruments these files).
2. Add source map support for `content-tag`
    Basically the code is first getting transpiled by `content-tag` from `gjs/gts` to `ts/js` and after that it is running through babel where the instrumentation happens, BUT when we transpile it with `content-tag` right now it doesn't generate source maps which are actually needed to map the transpiled code back to the original `gjs/gts` code, w/o this coverage lines will be off also it will report on some function that don't exist in the source code.
Created issues for this:
https://github.com/embroider-build/content-tag/issues/61
3. Remap the coverage data generated during test runs to the original code using [istanbul-lib-source-maps](https://github.com/istanbuljs/istanbuljs/tree/master/packages/istanbul-lib-source-maps) and providing it the `inputSourceMap` that we get from step 2.
This last step basically means we need to add some additional functionality to `ember-cli-code-coverage`, after that in theory it should work (I will work on this and try to make a PR, also will raise a separate issue for this part)

veelci:
[...]

PR #405, opened 2024-01-29
Add support for gjs/gts files

vstefanovic97:
> **Disclaimer:** Disclaimer this is just to make the js/ts part of the code have coverage, it does not support covering hbs part of the code. this is done in https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/pull/319

So to sum it up the steps in this PR are:

1. Add support for inputSourceMaps (remap coverage using https://github.com/istanbuljs/istanbuljs/tree/master/packages/istanbul-lib-source-maps)

When working with already transpiled code, that has sourcemaps included (inline or in separate file) in order for the lines to be correct we actually need to use [istanbul-lib-source-maps](https://github.com/istanbuljs/istanbuljs/tree/master/packages/istanbul-lib-source-maps) to remap the coverage to the correct source code.

This is actually needed since when we build v2 addons we already transpile the code from `gjs/gts` to `ts/js` first and we run babel on the already transpiled code, so we need to remap the coverage to the original source code.
Disclaimer: gts/gjs->js/ts transpilation doesn't generate sourcemaps at the moment but there is a PR open which will enable it https://github.com/embroider-build/content-tag/pull/62

2. Make sure to ignore template's `eval` method from the coverage report

Because we instrument already transpiled code we end up instrumenting some method that shouldn't be instrumented, like for example

```gjs
const var baz = <template>
  <button {{on "click" bar}}>
   increment
  </button>
</template>
```

gets instrumented to:

```js
var baz = template(`
  <button {{on "click" bar}}>
   increment
  </button>
`, {
  eval() {
    cov_1sbgmdzd2v().f[2]++; // we don't want istanbul to track this
    cov_1sbgmdzd2v().s[5]++; // we don't want istanbul to track this
    return eval(arguments[0]);
  }
});
```
Because istanbul instruments the `eval` method it will report as uncovered in tests and also it skews coverage percentages, so we want it to not instrument this, for this I created a small babel plugin that will add istanbul ignore lines and run before istanbuls babel plugin.

3. Include `.gjs/.gts` extensions in istanbul `extension` option by default so instrumentation will work on these files

4. Write tests (create a v2 addon with gjs/gts where we build it with coverage and then tests the coverage with snapshots)
  Lines in the snapshots coverage will still be offset until https://github.com/embroider-build/content-tag/pull/62 is fixed and 
  we update the test app with the change


vstefanovic97:
@kategengler seems like the error reported in CI was a known PNPM issue https://github.com/pnpm/pnpm/issues/5056 which was fixed with the latest pnpm version. I updated the version for CI and Volta do you mind to rerun the workflow and see if it fixes it?

vstefanovic97:
@RobbieTheWagner can you or maybe some other maintainer please review this?

test-packages/my-embroider-app-test.mjs
```diff
@@ -22,6 +22,7 @@ describe('app coverage generation', function () {
         "@embroider/webpack": embroiderVersion
       });
   
+      await execa('rm',  ['-rf', '.embroider'], { cwd: buildPath, env });
```
RobbieTheWagner, line 25: Can you please explain why this was needed?

vstefanovic97: @RobbieTheWagner  so currently we are testing against `emroider/core` versions from 0 to 3.
During each build it creates a `.embroider` folder, but seems like when we run a test for a newer version it will try to re-use the `.embrodier` folder generated by the previous test and it was leading to some unpredictability.

For example there is a problem with [babel plugin serialization](https://github.com/embroider-build/embroider/issues/1525) in v3 that I solved [here](https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/pull/405/files#diff-220612c243d58fc738e2512fd4efbd91e7700e2b8b547032a3e260df5fb7ee9dR92) using [this](https://github.com/Serabe/ember-bind-helper/pull/21) as a reference, but it would still fail on tests w/o cleaning up `.embroider` because it was relying on what was generated by the previous embroider version

So basically all I did was make sure the tests are totally isolated from each other and don't affect each other

SergeAstapov: @vstefanovic97 that's good knowledge! Mind to drop a comment in the code with this explanation for future readers?

vstefanovic97: Added a comment in the code about this @SergeAstapov, @RobbieTheWagner

vstefanovic97:
@RobbieTheWagner sorry to bother you again, any progress with the review?
Is there anything I can do to help make the reviewing easier, or anything additional that need explaining, any concerns etc. ?

RobbieTheWagner:
@vstefanovic97 I really don't have enough knowledge about how this addon works to review. @SergeAstapov do you feel comfortable reviewing?

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - uses: pnpm/action-setup@v2
         name: Install pnpm
         with:
-          version: 8.10.3
+          version: 8.15.1
```
SergeAstapov, line 23: @vstefanovic97 is this necessary for this PR? Can we do this bump separately?

packages/ember-cli-code-coverage/index.js
```diff
@@ -48,19 +58,25 @@ module.exports = {
       if (config.coverageEnvVar) {
         coverageEnvVar = config.coverageEnvVar;
       }
+
+      if (config.extension) {
```
SergeAstapov, line 62: @vstefanovic97 what is the use case when we want to configure this?
Also, as this exposes new option for consumers, let's document it in readme

packages/ember-cli-code-coverage/index.js
```diff
@@ -72,7 +88,12 @@ module.exports = {
     }
 
     const IstanbulPlugin = require.resolve('babel-plugin-istanbul');
-    return [[IstanbulPlugin, { cwd, include: '**/*', exclude }]];
+    return [
+      useStringLookupForPlugin
```
SergeAstapov, line 92: @vstefanovic97 could you please explain this ternary? can't we do e.g. `require` all the time?

package.json
```diff
@@ -8,6 +8,7 @@
     "test": "vitest"
   },
   "devDependencies": {
+    "@babel/core": "^7.23.9",
```
SergeAstapov, line 11: @vstefanovic97 why is this needed in the root of monorepo? I guess each package that needs it should declare such kind of dependency.

packages/ember-cli-code-coverage/lib/attach-middleware.js
```diff
@@ -134,17 +137,22 @@ function adjustCoverage(coverage, options) {
       namespaceMappings,
       modifyAssetLocation
     );
-    coverage[filePath].path = path.relative(root, relativeToProjectRoot);
-    memo[path.relative(root, relativeToProjectRoot)] = coverage[filePath];
+    coverage[filePath].data.path = path.relative(root, relativeToProjectRoot);
```
SergeAstapov, line 140: @vstefanovic97 out of curiosity, is this change needed because we now work with struct returned by `sourceMapStore.transformCoverage`?

packages/ember-cli-code-coverage/lib/gjs-gts-istanbul-ignore-template-plugin.js
```diff
@@ -0,0 +1,47 @@
...
+    visitor: {
+      Program: {
+        enter(path, state) {
+          const filename = state.file.opts.filename;
+
+          if (!filename?.match(/\.g[tj]s$/)) {
+            return;
+          }
+
+          path.traverse(gjsGtsTemplateIgnoreVisitor, state); // we need to do early traverse to make sure this runs before istanbuls plugin
```
SergeAstapov, line 42:
```suggestion
          // we need to do early traverse to make sure this runs before istanbuls plugin
          path.traverse(gjsGtsTemplateIgnoreVisitor, state);
```
just a tiny bit more readable

packages/ember-cli-code-coverage/lib/gjs-gts-istanbul-ignore-template-plugin.js
```diff
@@ -0,0 +1,47 @@
+const gjsGtsTemplateIgnoreVisitor = {
```
SergeAstapov, line 1: @vstefanovic97 let's provide explanation why we add this, which problem it solves and which other options e have considered but didn't go and why.

test-packages/my-v2-addon-gjs-gts/addon/babel.config.cjs
```diff
@@ -0,0 +1,25 @@
...
+    require('decorator-transforms'),
+    [
+      'babel-plugin-ember-template-compilation',
+      {
+        targetFormat: 'hbs',
+        transforms: [],
+      },
+    ],
+    ...require('ember-cli-code-coverage').buildBabelPlugin({
+      extension: ['.gjs', '.gts', '.js', '.ts', '.mts', '.cts', '.cjs', '.mjs'],
```
SergeAstapov, line 22: @vstefanovic97 this matches the default value so we can probably skip this?

test-packages/my-v2-addon-gjs-gts/addon/rollup.config.mjs
```diff
@@ -0,0 +1,72 @@
...
+    babel({
+      extensions: ['.js', '.gjs', '.ts', '.gts'],
+      babelHelpers: 'bundled',
+    }),
+
+    // Ensure that standalone .hbs files are properly integrated as Javascript.
+    addon.hbs(),
+
+    // Ensure that .gjs files are properly integrated as Javascript
+    addon.gjs({ inline_source_map: true }),
```
SergeAstapov, line 55: @vstefanovic97 looks like it's worth to document this option in readme.
Maybe, I'm thinking we need a section with steps how to setup test coverage in v2 addons.

SergeAstapov (changes requested):
@RobbieTheWagner Thank you for pinging me!
Functionally, this looks great to me! I'd suggest we spend a bit more effort on documentation part to make clear for ppl who'd like to start using this and for ppl who would need to understand code in future

packages/ember-cli-code-coverage/index.js
vstefanovic97: @SergeAstapov I can't think of a use case right now, but thought maybe it would be useful to expose it still, I'll add it to the docs

packages/ember-cli-code-coverage/index.js
vstefanovic97: There is a problem with [babel plugin serialization](https://github.com/embroider-build/embroider/issues/1525) in v3, I solved it using string lookups since I found [this](https://github.com/Serabe/ember-bind-helper/pull/21) as a reference.
I tried to use string lookup all the time, but it didn't work for v1 and v2

package.json
vstefanovic97: It's used in tests [here](https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/pull/405/files#diff-11851204535d7735df82299a725e271fc96f8c040a5468c6070cba4a7ce479f6R2)

packages/ember-cli-code-coverage/lib/attach-middleware.js
vstefanovic97: yes @SergeAstapov

.github/workflows/ci.yml
vstefanovic97: @SergeAstapov it's necessary, see https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/pull/405#issuecomment-1916394358

package.json
SergeAstapov: ah, now I see - `test-packages` folder contains both vite tests *and* test packages.
IMO seems confusing but for sure we should follow existing structure in this PR.

packages/ember-cli-code-coverage/index.js
vstefanovic97: @SergeAstapov I added this option to the docs

packages/ember-cli-code-coverage/index.js
vstefanovic97: @SergeAstapov Added comment about this to the code

packages/ember-cli-code-coverage/lib/gjs-gts-istanbul-ignore-template-plugin.js
vstefanovic97: Fixed @SergeAstapov

packages/ember-cli-code-coverage/lib/gjs-gts-istanbul-ignore-template-plugin.js
vstefanovic97: @SergeAstapov I added an explanation to the file

vstefanovic97:
@RobbieTheWagner looks like I had deleted something by accident, I fixed it now, mind running again? This time I think it will passf

README.md
```diff
@@ -154,6 +154,8 @@ Configuration is optional. It should be put in a file at `config/coverage.js` (`
 
 - `excludes`: Defaults to `['*/mirage/**/*']`. An array of globs to exclude from instrumentation. Useful to exclude files from coverage statistics.
 
+- `extension`: Defaults to `['.gjs', '.gts', '.js', '.ts', '.cjs', '.mjs', '.mts', '.cts']`. Tell Istanbul to instrument only files with the provided extensions.
```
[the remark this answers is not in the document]
vstefanovic97: @SergeAstapov well in the code it really is `extension` reason for this is I decided to keep it the same as the option in Isntabul since it too is singular

vstefanovic97:
@RobbieTheWagner I fixed latest comments from @SergeAstapov , mind to run CI again, neither I nor @SergeAstapov have permissions to run the CI :/

SergeAstapov (approved):
if you ask me, this looks great! Thank you @vstefanovic97 for working on this! cc @RobbieTheWagner @kategengler

RobbieTheWagner (approved):
If @SergeAstapov says this looks good, that is good enough for me 👍

vstefanovic97:
@RobbieTheWagner mind to run again (I think this will be the final time)? I changed test name but forgot to update the snapshot

SergeAstapov (approved):

vstefanovic97:
@RobbieTheWagner, @SergeAstapov can we merge this and create a new minor release?

SergeAstapov:
IMO yes, we are good to go, but I don't have permissions here.
@RobbieTheWagner would you be able to push this over the line and do the release?

cah-brian-gantzler:
After updating to this version I am now getting no coverage directory being created at all. Anything else I might have to do to configure? I did already have all the configurations from the readme in place for the prior version.

vstefanovic97:
@cah-brian-gantzler there should be no additional config required for it to work.

Do you see any Errors or something?
Is it a regular Ember app/addon, is it Embroider App or v2 addon?

Also if you can come up with a minimal reproduction I could also try to debug it

cah-brian-gantzler:
was on 2.0.3
No do not see any errors. 
it is an embroider app

Since its in my app, doubt and will prob work with a new app, not sure where to start on a minimal not knowing what the problem is.  

Will experiment some. Its weird that it produces NO coverage at all though. Will also revert back and make sure it still produces then. Did last time I asked, but that was a couple weeks ago, was waiting for this update

cah-brian-gantzler:
@vstefanovic97 must have been a glitch. Now produces COVERAGE directory, however no .gjs or .gts files appears appear (same as before)

I checked to ensure I had the correct package in my `node_modules`, and the package.json says `2.1.1` However noticed that the `changelog.md` is out of date. You might want to consider copying the `changelog.md` from the root to the package before publishing. 

I found I also picked up ` useBabelInstrumenter: true,` in the config file over the years. Not sure where I got that, but removing it did not alter the output.

Lastly I have my `mirage` directory in the `app` folder and not the root. Altering the excludes to `*/app/mirage/**/*` or `app/mirage/**/*` did not remove mirage from the report, but `**/mirage/**/*` did although that seems overkill. Got a suggestion on what the glob should be? Not great at globs

vstefanovic97:
@cah-brian-gantzler can you build the app with `COVERAGE=true` it can be either a production or development build, and check if the code for your gjs/gts files is instrumented with Istanbul?

Here is an example of what istanbul should do to the code, if you aren't familiar
https://gist.github.com/robertknight/834452c3b06963ff2a8b9682fd4189cb

Maybe also it's worth checking if you have source maps turned on for your tests
https://github.com/ember-template-imports/ember-template-imports?tab=readme-ov-file#sourcemap-generation
This part is actually needed for the lines to be accurate.

Let's first check this then we can narrow it down further.
I don't have time today, but I'll try this out on a new embroider app to see if it works.

Also I got some feedback here that it works https://github.com/ember-cli-code-coverage/ember-cli-code-coverage/issues/392#issuecomment-1959684127 so maybe what you are seeing is some edge case, but we will try to figure it out and fix :D

vstefanovic97:
@cah-brian-gantzler this looks to be an embroider + ember-cli-code-coverage with gjs/gts issue

I was able to reproduce with a band new embroider app.
I am still trying to dig into the root cause and give it a proper fix

For now can you try to add
```
const path = require('path');

module.exports = {
  modifyAssetLocation(root, relativePath) {
    if (relativePath.includes('/my-embroider-test/')) {
      return path.join('app', relativePath.split('/my-embroider-test/')[1]);
    }

    return false;
  },
};
```
to your `config/coverage.js` file
I think this will get the files to show, but for some reason coverage percentage is still not accurate (I'm still debugging this).
Btw do you mind opening an issue for this, it will be simpler to track there and not in this PR as it's closed

packages/ember-cli-code-coverage/index.js
```diff
@@ -72,7 +88,13 @@ module.exports = {
     }
 
     const IstanbulPlugin = require.resolve('babel-plugin-istanbul');
-    return [[IstanbulPlugin, { cwd, include: '**/*', exclude }]];
+    return [
+      // String lookup is needed to workaround https://github.com/embroider-build/embroider/issues/1525
```
NullVoxPopuli, line 92: As it turns out... We're going to require string lookup eventually, and forbid passing JS references.

JS references are not parallizable, but string references are! (it's a babel thing :shrug: )

NullVoxPopuli: one thing you could try is using a `resolve`d string, so you get the absolute path of the plugin for "this instance of ember-cli-code-coverage".

```js
path.resolve('the-string-you-have'); // returns absolute string, so no matter who is requiring, their resolve isn't local to who is doing the requiring 
```

PR #419, opened 2024-08-14
Add Config per Reporter

gossi:
I'm using codeclimate for my OSS projects. And for that to work, the code-coverage must be relative to repo-root, instead of project-root. This was easy done with the `vitest` project in there as it allowed to configure reports: https://vitest.dev/guide/reporters.html (it's also the same syntax that eslint uses to configure your rules).

In `ember-cli-code-coverage`, the config is always `{}` or "no config allowed". Now this PR adds support for the same config, that `vitest` allows.

I'm already using this with a patch on my OSS projects, here is one such [PR pointed to the config](https://github.com/hokulea/aria-voyager/pull/8/files#diff-1eac964ac8ee654baaf6c1c509383e0400a49292fda5a7b13e59249939159014).

It works, I can see my coverage on codeclimate and this PR brings back this support to everybody :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,13 @@ Configuration is optional. It should be put in a file at `config/coverage.js` (`
 
 - `coverageEnvVar`: Defaults to `COVERAGE`. This is the environment variable that when set will cause coverage metrics to be generated.
 
-- `reporters`: Defaults to `['lcov', 'html']`. The `json-summary` reporter will be added to anything set here, it is required. This can be any [reporters supported by Istanbul](https://github.com/gotwarlost/istanbul/tree/master/lib/report).
+- `reporters`: Defaults to `['lcov', 'html']`. The `json-summary` reporter will
+  be added to anything set here, it is required. This can be any [reporters
+  supported by
+  Istanbul](https://github.com/gotwarlost/istanbul/tree/master/lib/report).
+  Reporters can be configured with array-style syntax, for example, here are
+  options to `lcov` with a different `projectRoot`: `[['lcov', { projectRoot:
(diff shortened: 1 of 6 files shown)


RobbieTheWagner:
This is great, thanks @gossi! I spent the last few days fighting with Istanbul and nyc trying to get all my paths to be the same and ultimately settled on just manually modifying them with `sed`. I think this might fix that issue? 

Would it be possible to document a bit more about what `projectRoot` is or the use cases here? Like does `projectRoot` get prepended to the already determined paths? How is it different than passing `cwd` to the options etc

gossi:
Thanks @RobbieTheWagner but I must confess, I only copy pasted things around to make my use-case work 🙈 In my project I was able to configure it for `vitest`, that worked and then I made it possible to have the same config for `ember-cli-code-coverage`, which then also made it work (I was happy to ignore how that worked and not question it). So the combo `lcov` + `projectRoot` is what people might likely be interested, so that's enough of a hint.

On the other hand it doesn't matter that much for that PR. Since this is very generic to have _any_ options for _any_ reporter. So it is up to consumers to figure out the options for each reporter.

kategengler (approved):

RobbieTheWagner (approved):

RobbieTheWagner:
Thanks @gossi! I still think it would be nice to have some more docs about this and when to use it etc. but I do want to go ahead and get it in to unlock this.

jrjohnson:
Just wanted to throw some thanks on this, I was searching for how to configure the `projectRoot` and am thrilled to see it's already here! 🤩

